PR #99, opened 2019-03-21
1) New: add UIDialogController.wrapContentInScrollView for case where…

melsomino:
… dialog content contains list view.

2) New: add UIDialogController.getBottomPanelHeight which can be used in content list views.

### DIFF
--- a/controllers/UIDialogController/index.js
+++ b/controllers/UIDialogController/index.js
@@ -6,7 +6,6 @@ import {
     Text,
     StyleSheet,
     Animated,
-    Platform,
 } from 'react-native';
 
 import UIController from '../UIController';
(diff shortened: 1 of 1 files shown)


controllers/UIDialogController/index.js
```diff
@@ -313,6 +338,7 @@ class UIDialogController extends UIController {
                         styles.scrollContainer,
                         this.getScrollContainerStyle(),
                     ]}
+                    contentInset={contentInset}
```
sertony: It works only on iOS, padding should be applied for contentContainerStyle

controllers/UIDialogController/index.js
```diff
@@ -20,8 +19,6 @@ import UIDialogTextInput from './UIDialogTextInput';
 const styles = StyleSheet.create({
     scrollContainer: {
         justifyContent: 'center',
-        paddingTop: UIConstant.normalContentOffset(),
```
sertony: Top padding should be returned, otherwise all content will be sticked to the top

PR #105, opened 2019-03-29
Calendar fix locale

jcma86:
Fixed issue related to locale date when language is other than English
Reported by Alexey: [https://www.notion.so/tonlabs/Calendar-element-works-wrong-on-Web-and-Android-48e6e2905d064ab48931a94c73fd4019#48e6e2905d064ab48931a94c73fd4019](url)

### DIFF
--- a/components/text/UIDateInput/index.js
+++ b/components/text/UIDateInput/index.js
@@ -6,7 +6,6 @@ import type { KeyboardType } from 'react-native/Libraries/Components/TextInput/T
 
 import UIDetailsInput from '../UIDetailsInput';
 
-import UIStyle from '../../../helpers/UIStyle';
 import UIColor from '../../../helpers/UIColor';
 import UITextStyle from '../../../helpers/UITextStyle';
 import UIConstant from '../../../helpers/UIConstant';
--- a/components/text/UIDateInput/index.js
+++ b/components/text/UIDateInput/index.js
@@ -74,7 +74,7 @@ export default class UIDateInput extends UIDetailsInput<Props, State> {
         const { initialEpochTime } = this.props;
 
         if (initialEpochTime) {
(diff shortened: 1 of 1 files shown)


components/text/UIDateInput/index.js
```diff
@@ -112,15 +111,21 @@ export default class UIDateInput extends UIDetailsInput<Props, State> {
...
+        const validDate = Moment(date, this.getPattern(true)).isValid();
         const validLength = date.length === UIConstant.shortDateLength() || date.length === 0;
 
         return (validDate && validLength);
     }
 
-    getPattern() {
+    // This method returns a localized string that will be rendered
+    // in order to indicate the date component that has to be entered.
+    // However, internally (programming language), in order to parse/localized
```
sertony:
```suggestion
    // However, internally (programming language), in order to parse/localize
```

components/text/UIDateInput/index.js
```diff
@@ -112,15 +111,21 @@ export default class UIDateInput extends UIDetailsInput<Props, State> {
...
+    // a date, patter symbols have to be provided in English.
```
sertony:
```suggestion
    // a date, pattern symbols have to be provided in English.
```

components/text/UIDateInput/index.js
```diff
@@ -112,15 +111,21 @@ export default class UIDateInput extends UIDetailsInput<Props, State> {
...
+    getPattern(internalPattern: boolean = false) {
```
sertony: Maybe to change the naming? since it's not very clear what is "internal"...
I would suggest to add another opposite parameter called `localizedPattern` and use it in those methods which don't require **Moment**, it'd be much easier to read as I think...

PR #107, opened 2019-03-29
1) Fix: UIDialogController now use .Resize style on Android.

melsomino:


controllers/UIController/index.js
```diff
@@ -236,27 +245,49 @@ export default class UIController<Props, State>
...
     onKeyboardWillShow(event: KeyboardEvent) {
-        const end = event.endCoordinates;
+        const keyboardFrame = event.endCoordinates;
         const animation = UIController.getKeyboardAnimation(event);
         const { container } = this;
-        if (!(container && container.measure)) {
-            this.setBottomInset(end.height, animation);
-            return;
+        if (!!container && !!container.measure) {
+            this.adjustBottomInset(container, keyboardFrame, animation, 300);
```
sertony: I suggest to leave a comment for `300`, OR move this constant above and provide some proper method.

PR #1223, opened 2021-08-20
Use JSIExecutorInitializer to inject haptic functions

savelichalex:
Hermes has a check in debug mode for execution in runtime,
a call to it's methods should be on the same thread it was created.
If we tried to bind JSI methods from a RN native module it would be
on a different thread, therefore we have to use RCTCxxBridgeDelegate
method to set up them properly. Fortunatelly for us at the time
`react-native-reanimated` also supported this and we can safely use
it, but it requires us to have `reanimated` version no less than
`2.3.0-alpha.2`.


sertony:
Closing in favour of https://github.com/tonlabs/UIKit/pull/1250

PR #1225, opened 2021-08-23
Apply a guard only on ios

savelichalex:


packages/navigation/src/UILargeTitleHeader/useOnScrollHandler/onScroll.native.tsx
```diff
@@ -40,15 +42,14 @@ export default function (
...
-         * since the guard is needed only for iOS
+         * The fix is needed only for iOS
          *
          * On iOS `onScroll` event could fire on mount sometimes,
          * that's likely a bug in RN or iOS itself.
          * To prevent changes when there wasn't onBeginDrag event
          * (so it's likely not an actual scroll) using a guard
          */
-        if (ctx != null && !ctx.scrollTouchGuard) {
+        if (isIOS && !ctx != null && !ctx.scrollTouchGuard) {
```
sertony: Maybe you wanted to do this? I'm not sure why have you added the leading `!` to `ctx != null` condition...
```suggestion
        if (isIOS && ctx != null && !ctx.scrollTouchGuard) {
```

sertony (changes requested):
Consider changing the base into `surf@6.3.0` if it's a bug to fix for beta

sertony:
Btw, what bug does it fix?

savelichalex:
> Btw, what bug does it fix?

So this fix is not needed on Android, but it brakes normal behaviour, as we call `onScroll` without passing a context from a gesture handler (and we can't get this context, moreover we shouldn't)

sertony (approved):

SAnatoliiS (approved):

sertony (approved):

PR #1234, opened 2021-08-25
Check that y coordinate in scroll event is a number for UILargeTitleHeader

savelichalex:


sertony (approved):
We need to cherry-pick it into 6.3.0

sertony:
I'm still confused why was it working for 6.2.0

SAnatoliiS (approved):

PR #1239, opened 2021-08-27
Removing UIToastMessage

ghost:


Example/types/index.d.ts
```diff
@@ -2,4 +2,4 @@ declare module '@tonlabs/uikit.legacy';
 declare module '@tonlabs/uikit.navigation_legacy';
 declare module '@tonlabs/uikit.components';
 declare module '@tonlabs/uikit.core';
-declare module 'react-navigation-surf';
+declare module 'react-navigation-surf';
```
savelichalex: It actually should be already completely removed, where you met this? :)

packages/browser/src/UIBrowserList.tsx
```diff
@@ -115,30 +118,32 @@ const renderBubble = () => (
...
             [messages],
         );
         return (
-            <UrlPressHandlerContext.Provider value={onPressUrl}>
-                <UICommonChatList
-                    forwardRef={ref}
-                    nativeID="browserList"
-                    renderBubble={renderBubble()}
-                    getItemLayoutFabric={flatListGetItemLayoutFabric}
+            <TextLongPressHandlerContext.Provider value={onLongPressText}>
```
savelichalex: Together with @sertony we actually thought that it's might be good to move the context inside `UICommonChatList` to not duplicate it for `UIChatList` :)

packages/chats/src/UIChatList.tsx
```diff
@@ -99,29 +101,31 @@ export const UIChatList = React.forwardRef<SectionList, UIChatListProps>(
...
 
         return (
-            <UrlPressHandlerContext.Provider value={onPressUrl}>
-                <UICommonChatList
-                    forwardRef={ref}
-                    nativeID={nativeID}
-                    renderBubble={renderBubble}
-                    getItemLayoutFabric={sectionListGetItemLayout}
-                    canLoadMore={canLoadMore}
+            <TextLongPressHandlerContext.Provider value={onLongPressText}>
```
savelichalex: Yep, that what I'm talking about :)

packages/chats/src/UICommonChatList.tsx
```diff
@@ -424,6 +425,16 @@ type UICommonChatListProps<ItemT extends BubbleBaseT> = {
     canLoadMore?: boolean;
 };
 
+
+export const TextLongPressHandlerContext = React.createContext
```
savelichalex: Why we have two contexts? 🤔

packages/browser/src/UIBrowserList.tsx
```diff
@@ -126,24 +125,23 @@ export const UIBrowserList = React.forwardRef<FlatList, UIBrowserListProps>(
...
         );
         return (
-            <TextLongPressHandlerContext.Provider value={onLongPressText}>
-                <UrlPressHandlerContext.Provider value={onPressUrl}>
-                    <UICommonChatList
-                        forwardRef={ref}
-                        nativeID="browserList"
-                        renderBubble={renderBubble()}
-                        getItemLayoutFabric={flatListGetItemLayoutFabric}
+            <UrlPressHandlerContext.Provider value={onPressUrl}>
```
savelichalex: Could you please move `UrlPressHandlerContext` too? 😊

Example/types/index.d.ts
ghost: @savelichalex I don't know, it just was here. Should I remove this now?

savelichalex: Yep, definitely :) It's actually moved to `uikit.navigation` few month ago :)

ghost: @savelichalex Okay!

savelichalex (approved):
Great work, thanks!

packages/chats/src/BubblePlainText.tsx
```diff
@@ -153,24 +161,26 @@ function PlainTextContainer(
     const roundedCornerStyle = useBubbleRoundedCornerStyle(props, position);
     const actionString = getActionString(props);
 
+    const textLongPressHandler = useTextLongPressHandler()
+
+    const copyToClipboard = React.useCallback(()=>{
+        bubbleScaleAnimation(true);
+        props.text && Clipboard.setString(props.text);
+        textLongPressHandler && textLongPressHandler()
```
sertony: Not sure if we need to copy the text just right here (why not doing it from the parent?)
My thoughts is that you won't understand that the text was copied after the long press (since no toast has been shown yet about the fact it was copied...)

@savelichalex what do you think?

Also, I do think that we need to pass the text itself to the `textLongPressHandler` (as we do it for URL to process it in the parent)

packages/chats/src/BubblePlainText.tsx
```diff
@@ -153,24 +161,26 @@ function PlainTextContainer(
...
-                     * but I don't want to put it in legacy package
-                     * so left it here, until we make new share manager
-                     */
-                    runOnUI(hapticImpact)('medium');
-                    UIShareManager.copyToClipboard(
-                        props.text,
-                        uiLocalized.MessageCopiedToClipboard,
-                    );
-                }}
+                onLongPress={copyToClipboard}
```
sertony: Btw, It's not necessary that we'd like to copy the text to clipboard since it's up to parent how to behave IMHO, maybe it's better to rename the function to something like `handleLongPress` or alike

packages/chats/src/types.ts
```diff
@@ -45,6 +45,11 @@ export enum ChatMessageType {
 export type OnPressUrl =
     | ((url: string, index?: number) => void | Promise<void>)
     | undefined;
+
+export type OnLongPressText =
+    | (() => void | Promise<void>)
```
sertony: Here it is, I suggest to pass the text itself here

packages/navigation_legacy/src/UIShareManager/index.js
```diff
@@ -19,14 +17,6 @@ export type ShareOptions = {
...
 
 export default class UIShareManager {
-    // Private
-    static copyToClipboard(message: string, success: string) {
-        if (!message) {
-            return;
-        }
-        Clipboard.setString(message);
-        UIToastMessage.showMessage(success);
-    }
```
sertony, line 29: Oh, btw. Will you also apply changes in Surf? As there are many screens which use this function (as it actually not "private" as the comment claims it to be :D)

Plus we are to make Chat & Browser support the new `onLongPressText` prop as well

Example/types/index.d.ts
sertony: Btw, I've also removed it in my `RN0.65` PR which is yet in draft :)

packages/navigation_legacy/src/UIShareManager/index.js
ghost: @sertony To be honest, I had no idea how this function is used as it is used on many screens in surf. In fact, now I need to either do my own copyToClipboard with a notification call for each screen, or do a shared copyToClipboard that use sharedNotice, or simply return this function without a notification, and just add Notices on the screens. Which option do you think is more appropriate?

sertony: Oh, you will only need to pass the `onLongPressText` handle to both `UIChatList` & `UIBrowser` and process it for these two cases only.

In that case I would copy the text & show the notice directly from Surf (using Clipboard & UINotice instance (probably, shared is enough).

savelichalex (approved):

sertony (approved):

PR #1243, opened 2021-09-01
UIlarge title header with velocity

savelichalex:


packages/navigation/src/UILargeTitleHeader/useScrollHandler/useOnEndDrag.tsx
```diff
@@ -0,0 +1,275 @@
...
+             * how we calculate a shift and how a platform see it.
+             *
+             * Velocity factor is choosen with an eye test:
+             *  1) on iOS 500 should be read as 5 * 100, where:
+             *     - 100 is a multiplier of velocity as iOS gives very little
+             *       velocity value, that results to a very fast decoy animation ending;
+             *     - 5 was choosen with eye test, just to make it feel
+             *       like a continuation of the original scroll view movement.
+             *  2) TODO: Android
+             *  3) TODO: Web
```
sertony: What about it?)

sertony (approved):

packages/navigation/src/UILargeTitleHeader/useScrollHandler/useOnEndDrag.tsx
```diff
@@ -0,0 +1,275 @@
...
+            /**
+             * If we got to the point that means
+             * that no initial velocity was applied to a movement.
+             * It means that we should just adjust a position
+             * to the nearest snap point.
+             */
+            onWithoutDeceleration() {
+                'worklet';
+
+                console.log('onWithoutDeceleration');
```
SAnatoliiS: What is the need for this `console.log`?

packages/navigation/src/UILargeTitleHeader/useScrollHandler/useOnEndDrag.tsx
```diff
@@ -0,0 +1,275 @@
...
+                        overshootClamping: true,
+                    });
+
+                    return;
+                }
+                /**
+                 * If we are somewhere in between of a large title header
+                 * trying to find a nearest edge and move there
+                 */
+                if (shift.value > (0 - largeTitleHeight.value + defaultShift.value) / 2) {
```
SAnatoliiS: You can make it simpler: `(defaultShift.value - largeTitleHeight.value) / 2` :)

packages/navigation/src/UILargeTitleHeader/useScrollHandler/useOnEndDrag.tsx
```diff
@@ -0,0 +1,275 @@
...
+                     * Just fun story about scrolling on Android:
+                     * When you scrolling with your finger imagine the moment
+                     * when you release it, you actually can be surprised, but
+                     * actually there could be a backward movement that your pad
+                     * of a finger can produce, when it released, so at the end
+                     * you can get slight velocity in backward direction.
+                     * In oreder to reduce it treat such values as the ones
+                     * without inertia.
+                     * We achive this by accepting velocity values only more than 0.6.
+                     */
```
sertony, line 118: 😅

packages/navigation/src/UILargeTitleHeader/useScrollHandler/useOnEndDrag.tsx
```diff
@@ -0,0 +1,275 @@
...
+                        // Nothing to do, regular scroll
+                        if (event.contentOffset.y > 0) {
+                            return;
+                        }
+
+                        movementHandlers.onToBottomDeceleration(
+                            -1 * ctx.lastApproximateVelocity,
+                            runOnUIPlatformSelect({
+                                android: 70,
+                                default: 50,
```
SAnatoliiS, line 133: Probably you need to put all the values in constants and use `// @inline`. This also applies to other modules.

savelichalex: I actually don't think that it have to be moved to constants as it's not gonna be re-used, actually it might be bad to re-use them, as it should be applied only here, and If someone was changed it, it would break the carefully chosen behaviour :) It's like you don't want to move second parameter of `setTimeout` to constants, until it's a common case :)

packages/navigation/src/UILargeTitleHeader/useScrollHandler/useOnEndDrag.tsx
savelichalex: I just forgot to delete 😅

SAnatoliiS (approved):

PR #1245, opened 2021-09-02
New UICarouselView

ghost:


yarn.lock
```diff
@@ -12410,6 +12410,11 @@ react-native-pager-view@5.1.9:
   resolved "https://registry.yarnpkg.com/react-native-pager-view/-/react-native-pager-view-5.1.9.tgz#54dafe0ec459018c925ce319fb22df0f83d08d6c"
   integrity sha512-cwBhjhr+n+VI3UHOGJg2bq9GHO/KmMuke/QH5T+hPdnS/wems5g4J96A8c+EjRCLuUrU1uSS0UQO4ONQN+5kfg==
 
+react-native-pager-view@^5.4.1:
```
sertony: I’ve noticed that UIKit has installed two versions of pager here:
```
react-native-pager-view@5.1.9:
  version "5.1.9"
  resolved "https://registry.yarnpkg.com/react-native-pager-view/-/react-native-pager-view-5.1.9.tgz#54dafe0ec459018c925ce319fb22df0f83d08d6c"
  integrity sha512-cwBhjhr+n+VI3UHOGJg2bq9GHO/KmMuke/QH5T+hPdnS/wems5g4J96A8c+EjRCLuUrU1uSS0UQO4ONQN+5kfg==

react-native-pager-view@^5.4.1:
  version "5.4.1"
  resolved "https://registry.yarnpkg.com/react-native-pager-view/-/react-native-pager-view-5.4.1.tgz#88641ff083ca4379497dc03891fdc2d5b6f82263"
  integrity sha512-P/aGuJqRe/HaocYd8o6TsO9YQMFPQrVnkjph9vZ+yg3HCCF/7XLRlntOLEs24Vmg4M+02MjpH557WACpBADi/A==
  ```

File: packages/flask/src/UICarouselView/types.ts [hunk 2]
```diff
@@ -0,0 +1,47 @@
...
+export type CarouselChildren = React.ReactElement<UICarouselViewPageProps>[] | React.ReactElement<UICarouselViewPageProps>
+| React.ReactElement<UICarouselViewPageProps>[];
+/**
+ * Container properties.
+ */
+export type UICarouselViewContainerProps = {
+   /** callback that is called when the page changes */
+   onPageIndexChange?: (newPageIndex: number) => void;
+   /** determines whether pages should move when clicked */
+   isPageMovesOnPress?: boolean
```
sertony: This is a bit cleaner in English
```suggestion
   shouldPageMoveOnPress?: boolean
```

packages/flask/package.json
```diff
@@ -40,23 +40,24 @@
...
     "@types/react-native": "0.64.12",
-    "@tonlabs/babel-plugin-transform-inline-consts": "*",
+    "dayjs": "1.9.1",
     "react": "17.0.1",
     "react-dom": "17.0.1",
     "react-native": "0.64.2",
     "react-native-gesture-handler": "^1.10.3",
     "react-native-reanimated": "^2.2.0",
-    "react-native-svg": "^12.1.1",
-    "react-native-view-shot": "^3.1.2",
```
sertony: It looks like we still need `react-native-view-shot` in devDependencies (or I missing something?)

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/constants.ts
```diff
@@ -0,0 +1,6 @@
+import { Animated } from "react-native";
```
sertony: Do we need to take `.timing` for  react-native's Animated here? (Why it's not Reanimated?)

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.native.tsx
```diff
@@ -0,0 +1,106 @@
+import React from "react"
+import { throttle } from "lodash";
+import { View, StyleSheet, Pressable } from "react-native"
+import Animated from "react-native-reanimated";
+import PagerView from "react-native-pager-view";
+
+import type { UICarouselViewContainerProps, UICarouselViewPageProps } from "../../types";
+import { usePageStyle } from "../hooks";
+
+const returnPages = (pages:  React.ReactElement<UICarouselViewPageProps>[], isPageMovesOnPress: boolean,  nextPage: any) => {
```
sertony:
```suggestion
const returnPages = (pages:  React.ReactElement<UICarouselViewPageProps>[], shouldPageMoveOnPress: boolean,  nextPage: any) => {
```

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.native.tsx
```diff
@@ -0,0 +1,106 @@
+import React from "react"
+import { throttle } from "lodash";
```
sertony:
```suggestion
import throttle from "lodash/lodash";
```

Otherwise it will import the whole `lodash` which we don't need to have in the bundle. We only need the `throttle` module here.

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.native.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { View, StyleSheet, Pressable } from "react-native"
+import Animated from "react-native-reanimated";
+import PagerView from "react-native-pager-view";
```
sertony: Besides, AFAIR the quotes should be singular (`'` instead of `"`)

(At least it's true for Surf, if it's not in UIKit, I'll force the prettier rule it in my [RN0.65](https://github.com/tonlabs/UIKit/pull/1219) PR)

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.native.tsx
```diff
@@ -0,0 +1,106 @@
...
+        const ChildView = item.props.component;
+        const key = `UICarouselPage_${index}`;
+        return (
+            <Pressable
+                disabled={!isPageMovesOnPress}
+                key={key} 
+                testID={item.props.testID} 
+                onPress={throttle(() => 
+                            nextPage(index), 250, { leading: false, trailing: true}
+                        )}
```
sertony: Let's create a `useCallback` function for that? (for optimization purposes to avoid function recreation and ChildView re-render if it's Pure (i.e. React.memo))

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.native.tsx
```diff
@@ -0,0 +1,106 @@
...
+            `UICarouselViewContainer: children must have at least 1 item`,
+        );
+        return null;
+    }
+    
+    return(
+        <View testID={testID} style={styles.carouselViewContainer}>
+            <AnimatedPagerView 
+                ref={pagerRef}
+                style={[styles.carouselView, pageStyles]}
```
sertony: It looks weird that `pageStyles` is applied to the container itself (instead of the concrete page).
Or did you mean `pagerStyle`?

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.web.tsx
```diff
@@ -0,0 +1,283 @@
...
+    GestureResponderEvent,
+    PanResponderGestureState,
+    View,
+    LayoutChangeEvent,
+    Pressable
+} from "react-native"
+
+import type  { UICarouselViewContainerProps, UICarouselViewPageProps } from "../../types";
+import { useAnimatedValue } from "../hooks";
+import { AnimationConfig } from "./constants";
```
sertony: Oh, I see it now. You made it work on web using regular `Animated`...

To be honest I'd like to see `react-native-reanimated` here instead, especially because it works great on web starting version `2.x` (at the moment we use `2.2.0`).

sertony (changes requested):
We also need to discuss if we want to make it work on web using `reanimated` instead... (or a plain `react-native-pager-view`)

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.web.tsx
ghost: @sertony As far as I understand, the main task of Reanimated is to optimize native animation, but on the web this is common JS. Then why rely on reanimated in web when we can use stable React Native animations? 

Also I don't really understand how to implement this through reanimated, after all, I don't have much experience with him. Therefore, in order not to waste time, I used PanResponder from RN.

Example/package.json
```diff
@@ -70,7 +70,7 @@
         "react-native-lightbox": "0.7.0",
         "react-native-localization": "2.1.3",
         "react-native-navigation-bar-color": "2.0.1",
-        "react-native-pager-view": "5.1.9",
+        "react-native-pager-view": "^5.4.4",
```
savelichalex: Please "fix" the version (remove `^`) :)

package.json
```diff
@@ -41,6 +41,7 @@
         "react-native-gesture-handler": "^1.10.3",
         "react-native-image-picker": "*",
         "react-native-navigation-bar-color": "^2.0.1",
+        "react-native-pager-view": "^5.4.4",
```
savelichalex, line 44: @sertony I always forget, do we need to fix it there? 🤔

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.web.tsx
```diff
@@ -0,0 +1,189 @@
...
+    const jumpToIndex = React.useCallback(
+        (index: number) => {
+            const offset = -index * layout.width;
+            carouselOffset.value = withTiming(offset, { duration });
+            itemOffset.value = withSequence(
+                withTiming(1, { duration: duration / 2 }),
+                withTiming(0, { duration }),
+            );
+            setIsMoving(false);
+            onPageChangedRef.current && onPageChangedRef.current(index);
```
savelichalex: I actually didn't get what this optimization is about :) Can you please enlighten me a bit, what was the reasoning?

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.web.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+    React.useEffect(() => {
+        if (layoutRef.current.width && currentIndexRef.current !== initialIndex) {
+            currentIndexRef.current = initialIndex;
+            jumpToIndex(initialIndex);
+        }
+    }, [jumpToIndex, layoutRef.current.width, initialIndex]);
+
+    const handleLayout = React.useCallback((e: LayoutChangeEvent) => {
+        const { height, width } = e.nativeEvent.layout;
```
savelichalex:
```suggestion
    const handleLayout = React.useCallback(({ nativeEvent: { layout: { height, width } } }: LayoutChangeEvent) => {
```

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.web.tsx
```diff
@@ -0,0 +1,189 @@
...
+            return { height, width };
+        });
+    }, []);
+
+    return (
+        <View testID={testID} onLayout={handleLayout} style={[styles.pager]}>
+            <Animated.View
+                style={[
+                    styles.sheet,
+                    layout.width ? { width: maxWidthTranslate, ...containerAnimatedStyle } : null,
```
savelichalex, line 141:
```suggestion
                    layout.width ? [{ width: maxWidthTranslate }, containerAnimatedStyle ] : null,
```

packages/flask/src/UICarouselView/CarouselView/Pagination.tsx
```diff
@@ -0,0 +1,86 @@
...
+    onPress: (event: any) => void;
+    theme: Theme;
+};
+
+const Circle: React.FC<CircleProps> = ({ active, onPress, theme }: CircleProps) => {
+    const { animatedStyles } = usePaginationStyle(active, theme);
+
+    return (
+        <TouchableOpacity
+            hitSlop={{
```
savelichalex: Please extract it to the variable outside of the component

packages/flask/src/UICarouselView/CarouselView/Pagination.tsx
```diff
@@ -0,0 +1,86 @@
...
+                top: 8,
+                left: 8,
+                right: 8,
+                bottom: 8,
+            }}
+            onPress={onPress}
+            style={{ alignItems: 'center' }}
```
savelichalex: Can you put it into static styles? :)

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.web.tsx
ghost: @savelichalex No, it didn't work that way

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.web.tsx
ghost: @savelichalex In this case, saving to ref makes the callback more stable. I don't really understand what is the reason for lags without ref, because I did not delve into it.

package.json
sertony: @savelichalex it's done right here, since we don't need to fix `peerDependencies`

packages/flask/package.json
```diff
@@ -40,7 +40,9 @@
         "@tonlabs/uikit.localization": "^2.3.0",
         "@tonlabs/uikit.navigation": "^2.3.0",
         "html2canvas": "^1.0.0-rc.7",
-        "qrcode": "^1.4.4"
+        "lodash": "^4.17.21",
+        "qrcode": "^1.4.4",
+        "react-native-pager-view": "5.4.4"
```
sertony: This once should be moved to `devDependencies`! (Since it should be installed in the main project only as it contains native code, i.e. Example or Surf)

savelichalex: It's actually should be in `peerDependencies` if it contains native code :)

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.web.tsx
savelichalex: Oh, I guess it's because you didn't pass it to deps array for `useEffect`

savelichalex: Please try to remove it, as for now it looks like an unpredictable hack :)

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.web.tsx
```diff
@@ -0,0 +1,195 @@
...
+        currentIndex,
+        width,
+        offset,
+        onPress,
+        shouldPageMoveOnPress,
+        key,
+    }: PageProps) => {
+        const PageComponent = page.props.component;
+
+        const animatedStyles = useAnimatedStyle(() => {
```
sertony, line 27: Nice! Glad you switched to `reanimated` here ;)

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.web.tsx
sertony: Interesting.. also not very clear to me)

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.web.tsx
```diff
@@ -0,0 +1,195 @@
...
+    React.useEffect(() => {
+        layoutRef.current = layout;
+        const offset = -currentIndexRef.current * layoutRef.current.width;
+        if (onPageIndexChange) {
+            onPageChangedRef.current = onPageIndexChange;
+        }
+        carouselOffset.value = offset;
+    }, [layoutRef, carouselOffset, onPageIndexChange, layout]);
+
+    React.useEffect(() => {
```
sertony: Not very clear approach here, can you comment it?

packages/flask/src/UICarouselView/CarouselView/Pagination.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import { View, StyleSheet } from 'react-native';
+import Animated from 'react-native-reanimated';
+
+import { UIConstant } from '@tonlabs/uikit.core';
+import { Theme, TouchableOpacity, useTheme } from '@tonlabs/uikit.hydrogen';
+import { usePaginationStyle } from './animations';
+
+const circleSize = 6;
```
sertony: I always suggest to move constants into `constants` ;)

packages/flask/src/UICarouselView/CarouselView/Pagination.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+type CircleProps = {
+    active: boolean;
+    onPress: (event: any) => void;
+    theme: Theme;
+};
+
+const CircleHitSlop = {
```
sertony: Same

packages/flask/src/UICarouselView/CarouselView/UICarouselViewContainer.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { Pagination } from './Pagination';
+
+export const UICarouselViewContainer: React.FC<UICarouselViewContainerProps> = ({
+    testID,
+    children,
+    initialIndex = 0,
+    onPageIndexChange,
+    shouldPageMoveOnPress,
+    showPagination = true,
+}: UICarouselViewContainerProps) => {
```
sertony: as @savelichalex noted one day that it's better to use regular `function name(...args)` for components function to avoid anonymous functions in React DevTools.

Plus, there is another concern about it: https://twitter.com/heroicamit/status/1437677666929315840?s=12

Can you change it please?

packages/flask/src/UICarouselView/CarouselView/CarouselContainer/index.web.tsx
ghost: Managed to get rid of it

packages/flask/package.json
```diff
@@ -40,7 +40,8 @@
         "@tonlabs/uikit.localization": "^2.3.0",
         "@tonlabs/uikit.navigation": "^2.3.0",
         "html2canvas": "^1.0.0-rc.7",
-        "qrcode": "^1.4.4",
+        "lodash": "^4.17.21",
+        "qrcode": "^1.4.4"
         "fuse.js": "^3.4.6"
```
sertony:
```suggestion
        "qrcode": "^1.4.4",
        "fuse.js": "^3.4.6"
```

sertony (approved):

savelichalex (approved):

PR #1249, opened 2021-09-07
Add toast UINotice action.

SAnatoliiS:


packages/popups/src/Notice/Action.tsx
```diff
@@ -0,0 +1,27 @@
...
+                title={title}
+                onPress={onTap}
+            />
+        </View>
+    );
+};
+
+const styles = StyleSheet.create({
+    container: {
+        paddingLeft: 8,
```
savelichalex: Maybe use https://github.com/tonlabs/UIKit/blob/development/packages/hydrogen/src/constants.ts#L25 ?

savelichalex: Or the one from navigation

savelichalex (approved):

sertony (approved):

PR #1251, opened 2021-09-09
Add toast countdown

SAnatoliiS:


packages/popups/src/Notice/Notice.tsx
```diff
@@ -7,34 +7,50 @@ import {
...
+import { CountdownCirlce } from './CountdownCircle';
 
 const getBackgroundColor = (color: UINoticeColor, theme: Theme): ColorValue => {
     switch (color) {
         case UINoticeColor.Secondary:
             return theme[ColorVariants.BackgroundSecondary];
         case UINoticeColor.Negative:
-            return theme[ColorVariants.BackgroundNegative];
+            /**
+             * TODO Add new color to ColorVariants
```
savelichalex: Will you add it? :)

packages/popups/src/Notice/Notice.tsx
```diff
@@ -52,42 +68,51 @@ export const Notice: React.FC<NoticeProps> = ({
...
+                    onLongPress={onLongPress}
+                    delayLongPress={UIConstant.notice.longPressDelay}
+                    onPressOut={onPressOut}
                 >
-                    {title}
-                </UILabel>
+                    <View style={StyleSheet.absoluteFill} />
+                </TouchableWithoutFeedback>
+                <View style={styles.countdown}>
+                    <CountdownCirlce
```
savelichalex: So countdown isn't opt-in? 🤔

SAnatoliiS: Of course! I forgot to add the prop.

packages/popups/src/Notice/Notice.tsx
SAnatoliiS: When the designer adds this color to the palette

savelichalex: Can you please discuss it with Aleksei and let's add it now, no need to wait for it

SAnatoliiS: Done

packages/popups/src/Notice/types.ts
```diff
@@ -34,11 +33,15 @@ export type NoticeProps = {
     onLongPress: () => void;
     onPressOut: () => void;
     action?: UINoticeActionAttributes;
+    countdownValue: Animated.SharedValue<number>;
+    countdownProgress: Animated.SharedValue<number>;
+    isCountdownVisible?: boolean;
```
savelichalex: What do you think to call it `hasCountdown` ? It's shorter :)

savelichalex (approved):

packages/popups/src/Notice/CountdownCircle.tsx
```diff
@@ -0,0 +1,107 @@
...
+         * Used for web
+         */
+        'stroke-dashoffset': true,
+    }),
+);
+
+const circleOffset = 0.5;
+
+const size = 20;
+const strokeWidth = 1.2;
```
sertony: Why not moving it Constants?

sertony (approved):

PR #1258, opened 2021-09-12
RTL support to balance component added

NoamDev:
Balance component is shown in reverse on RTL locales. This PR is fixing this by using row-reverse when direction is RTL.


savelichalex:
@NoamDev I started to work on new balance there https://github.com/tonlabs/UIKit/pull/1262

savelichalex:
Close it in favour of https://github.com/tonlabs/UIKit/pull/1262

PR #1260, opened 2021-09-13
Fix UIMsgButton icon horizontal offset.

SAnatoliiS:


packages/hydrogen/src/UIMsgButton.tsx
```diff
@@ -366,92 +342,80 @@ export const UIMsgButton = ({
...
-                        }
-                    </Button.Content>
-                    {
-                        iconPosition === UIMsgButtonIconPosition.Right && icon != null &&
+            <>
+                <Button.Content>
+                    {iconPosition === UIMsgButtonIconPosition.Left && icon != null && (
                         <Button.Icon
                             source={icon}
+                            style={!!title ? styles.leftIcon : null}
```
savelichalex: So this is the fix, right? :)
Maybe extract to a variable and use this check instead in all places? (I like this check more than on 388 line) :)

sertony (approved):

savelichalex (approved):

PR #1262, opened 2021-09-14
New balance animation

savelichalex:


packages/flask/src/index.ts
```diff
@@ -3,5 +3,6 @@ export * from './UIAddressTextView';
 export * from './UIQRCodeView';
 export * from './UIDateTimePickerView';
 export * from './UIPinCode';
+export * from './UIAnimatedBalance';
```
SAnatoliiS, line 6: All new components should be added to Example

packages/flask/src/UIAnimatedBalance.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { uiLocalized } from '@tonlabs/uikit.localization';
+
+const AnimatedTextInput = Animated.createAnimatedComponent(TextInput);
+
+export function getRandomNum() {
+    const num = Math.random();
+    const symbols = 10 ** (Math.floor(Math.random() * 10) + 1);
+
+    return Math.floor(num * symbols) / 100;
+}
```
SAnatoliiS: I think this function should be moved to Example

packages/flask/src/UIAnimatedBalance.tsx
```diff
@@ -0,0 +1,112 @@
...
+            />
+        </View>
+    );
+}
+
+const styles = StyleSheet.create({
+    ltrContainer: {
+        flexDirection: 'row',
+    },
+    // On RTL RN will reverse `flex-direction` so to draw
```
sertony: 👍

sertony (commented):
Let's plan to switch on this one and remove the legacy one, once Surf migrates on it.

Plus we'll be able to close the following [PR](https://github.com/tonlabs/UIKit/pull/1258)

savelichalex:
> Let's plan to switch on this one and remove the legacy one, once Surf migrates on it.
> 
> Plus we'll be able to close the following [PR](https://github.com/tonlabs/UIKit/pull/1258)

Yep, we'll do, just have some design discussions and decisions going, stay in touch ;)

packages/flask/src/UIAnimatedBalance/UIAnimatedBalance.tsx
```diff
@@ -0,0 +1,286 @@
...
+                    <AnimatedTextInput
+                        style={[Typography[integerVariant], colorStyle, styles.integer]}
+                        animatedProps={animatedIntegerProps}
+                        defaultValue={formatted.value.integer}
+                        underlineColorAndroid="transparent"
+                        editable={false}
+                    />
+                    <Text style={[Typography[fractionalVariant], colorStyle, styles.delimeter]}>
+                        {decimalSeparator}
+                    </Text>
```
SAnatoliiS, line 234: In my opinion, the separator should not be displayed if `maxFractionalDigits === 0`

packages/flask/src/UIAnimatedBalance/UIAnimatedBalance.tsx
```diff
@@ -0,0 +1,286 @@
...
+            ) {
+                return;
+            }
+
+            const numRegExp = /[0-9]/g;
+            const wideNumLiteral = '5';
+
+            normalized.value =
+                state.integer.replace(numRegExp, wideNumLiteral) +
+                decimalSeparator +
```
SAnatoliiS, line 170: The same remark. What if `maxFractionalDigits === 0`?

packages/flask/src/UIAnimatedBalance/UIAnimatedBalance.tsx
```diff
@@ -0,0 +1,286 @@
...
+                    underlineColorAndroid="transparent"
+                    editable={false}
+                />
+                {/*
+                 * Space letter here is important here, it's not a mistake!
+                 * It's placed to have a separation from main balance,
+                 * instead of `marginLeft` as space is more "text friendly"
+                 * and will layout properly in text
+                 */}
+                <Text style={Typography[fractionalVariant]}> {iconElement}</Text>
```
SAnatoliiS, line 223: Will this white space work fine with RTL?

packages/flask/src/UIAnimatedBalance/UIAnimatedBalance.tsx
```diff
@@ -0,0 +1,286 @@
...
+            if (
+                prevState != null &&
+                state.integer.length + state.fractional.length ===
+                    prevState.integer.length + prevState.fractional.length
+            ) {
+                return;
+            }
+
+            const numRegExp = /[0-9]/g;
+            const wideNumLiteral = '5';
```
SAnatoliiS, line 166: The constants should probably be taken out?

savelichalex: I don't want them to be captured and turned into SharedValues :) especially RegExp, it works badly with reanimated when it's captured

packages/flask/src/UIAnimatedBalance/UIAnimatedBalance.tsx
savelichalex: Yep, text is rendered as usual in RTL, it's more about about an edge of container where it sticks

packages/flask/src/UIAnimatedBalance/UIAnimatedBalance.tsx
SAnatoliiS: Will it not properly work with the tag `@inline`?

packages/flask/src/UIAnimatedBalance/UIAnimatedBalance.tsx
savelichalex: Good point 🤔 I'm just thinking that what the point to show balance in blockchain when there is no fractional digits? we actually not making general thing to show any number, so to make it simpler, what if imply that there should be some amount of fractional digits? ie set type of `maxFractionalDigits` to sth like `1 | 2 | 3 | 4 | 5`, that would also eliminate case when someone passes huge count of fractional digits, that would be filled with 0

packages/flask/src/UIAnimatedBalance/UIAnimatedBalance.tsx
savelichalex: I did only numbers for now :)

SAnatoliiS: it's a pity, but not critical :)

packages/flask/src/UIAnimatedBalance/UIAnimatedBalance.tsx
SAnatoliiS: Maybe you're right. But I would be reinsured, and worked out this moment. We do not know where and for what purposes else this component can be used. It is possible that integer precision will be enough for someone or sometime for us.

savelichalex: For us, I know exact places where it's used, and it has 2 fractional digits in both cases (Initially I even thought to make it exactly 2, but then decided to give a bit of flexibility with a prop as it was before (and it wasn't used actually)). It's actually designed for this particular place, and to use it in different context may require another design, and fractional count might be the lesser problem, so I vote to solve problems as they appear :)

SAnatoliiS: I don't mind :)

SAnatoliiS (approved):

packages/flask/src/UIAnimatedBalance/UIAnimatedBalance.tsx
```diff
@@ -0,0 +1,286 @@
...
+    value,
+    maxFractionalDigits = 2,
+    icon,
+    loading = false,
+}: {
+    value: number;
+    icon: UIImageProps['source'];
+    integerVariant?: TypographyVariants;
+    fractionalVariant?: TypographyVariants;
+    maxFractionalDigits?: number;
```
savelichalex:
```suggestion
    maxFractionalDigits?: 1 | 2 | 3 | 4 | 5;
```

packages/flask/src/UIAnimatedBalance/UIAnimatedBalance.tsx
```diff
@@ -0,0 +1,286 @@
...
+    value,
+    maxFractionalDigits = 2,
+    icon,
+    loading = false,
+}: {
+    value: number;
+    icon: UIImageProps['source'];
+    integerVariant?: TypographyVariants;
+    fractionalVariant?: TypographyVariants;
+    maxFractionalDigits?: 1 | 2 | 3 | 4 | 5;
```
savelichalex, line 103:
```suggestion
    // We don't want to deal with edge cases for now
    // as the component is not for use for general cases with all number
    maxFractionalDigits?: 1 | 2 | 3 | 4 | 5;
```

vetkamak (approved):

PR #1263, opened 2021-09-15
Fix tintColor for Safari.

SAnatoliiS:


packages/hydrogen/src/UIImage.web.tsx
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Scaling is necessary in order to keep the image quality at the same level
+ */
+const scale = 2;
+
+const TintUIImage = ({ tintColor, ...rest }: UIImageProps) => {
+    const theme = useTheme();
+    const tintColorValue: ColorValue | null = tintColor != null ? theme[tintColor] : null;
+
+    const idRef = React.useRef(Math.random());
```
savelichalex: What do you think to use to use sth like `nanoid` to reduce possible collisions?

packages/hydrogen/src/UIImage.web.tsx
```diff
@@ -0,0 +1,108 @@
...
+            // set composite mode
+            ctx.globalCompositeOperation = 'source-in';
+            // draw color
+            ctx.fillStyle = tintColorValue as string;
+            ctx.fillRect(0, 0, width, height);
+        };
+        img.src = uri;
+    }, [uri]);
+
+    if (hasError || !tintColorValue || !width || !height || !uri) {
```
savelichalex: What do you think to add a warning log for developers (wrapped with `if (__DEV__)`) to warn that there is `tintColor` provided, but width and height aren't? :) (I saw places like this, and it might be not obvious for them that it might break sth :) )

packages/hydrogen/src/UIImage.web.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <View
+            style={{
+                width,
+                height,
+            }}
+        >
+            <View
+                style={[
+                    styles.tintImageContent,
+                    {
```
savelichalex: Maybe wrap it with `useMemo` or wrap the whole component with `React.memo`? :)

packages/hydrogen/src/UIImage.web.tsx
SAnatoliiS: The width and height do not have to be passed directly. If they were not passed in styles , it will be taken from the source. But I added the error to the console.

packages/hydrogen/src/UIImage.web.tsx
```diff
@@ -0,0 +1,161 @@
+/**
+ * Delete this file when the issue is fixed:
+ * https://github.com/necolas/react-native-web/issues/1914
+ */
+import * as React from 'react';
+import { nanoid } from 'nanoid';
```
savelichalex: Please ensure that it's set in `package.json` as a dependency :)

packages/hydrogen/package.json
```diff
@@ -53,6 +53,7 @@
         "@types/jest": "26.0.24",
         "@types/react": "17.0.20",
         "@types/react-native": "0.65.0",
+        "nanoid": "^3.1.23",
```
savelichalex: I don't think it's a devDep, rather usual a dep 🤔

packages/hydrogen/src/UIImage.web.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+import { ColorValue, Image as RNImage, ImageProps, StyleSheet, View } from 'react-native';
+import { ColorVariants, useTheme } from './Colors';
+
+const useImageDimensions = (style: any, source: any) => {
+    return React.useMemo(() => {
+        let width = 0;
+        let height = 0;
+        if (style) {
+            if (style.width) {
```
savelichalex: I'm curios how it actually works for web, AFAIK it doesn't return an object, that's why i usually use `StyleSheet.flatten` for the purpose :)

packages/hydrogen/src/UIImage.web.tsx
```diff
@@ -0,0 +1,161 @@
...
+                onLoadEnd();
+            }
+        };
+        img.src = uri;
+    }, [uri]);
+
+    if (hasError || !tintColorValue || !width || !height || !uri) {
+        if (__DEV__) {
+            console.error(
+                `UIImage.web.tsx: canvas rendering error, tintcolor will not be applied.${
```
savelichalex:
```suggestion
                `UIImage.web.tsx: there was tintColor provided for image, but no width and height specified, please ensure that they're passed correctly.${
```
Maybe sth like this? :)

SAnatoliiS: I took this, but made some corrections.

savelichalex (approved):

packages/hydrogen/src/UIImage.web.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    if (tintColor) {
+        return <TintUIImage {...props} />;
+    }
+
+    return React.createElement(RNImage, rest);
+};
+
+/**
+ * Delete this file when the issue is fixed:
```
sertony: Btw, since we have `.web.tsx` file we can leave it, but instead stop using `ImageComponent` in a regular `.tsx` file and instead convert it to `.native.tsx` file and use  `react-native-fast-image` there only without any `Platform.OS === 'web` check

packages/hydrogen/src/UIImage.web.tsx
```diff
@@ -0,0 +1,155 @@
...
+            return;
+        }
+        img.onload = () => {
+            if (!ctx || !dimensions) {
+                return;
+            }
+            // draw image
+            ctx.drawImage(img, 0, 0, dimensions.width, dimensions.height);
+            // set composite mode
+            ctx.globalCompositeOperation = 'source-in';
```
sertony: Oh, I'm starting to recall such "blend"mode names (was useful when creating the photo editor) 😄👍

sertony (approved):

PR #1265, opened 2021-09-15
Fix pan size

ghost:


packages/hydrogen/src/UISwitcher/ToggleSwitcher/index.tsx
```diff
@@ -33,46 +27,41 @@ export const ToggleSwitcher: React.FC<UISwitcherProps> = ({
...
-                    style={[styles.toggleInnerStyle, !disabled && overlayStyle]}
-                >
-                    <PanGestureHandler onGestureEvent={panGestureHandler}>
-                        <Animated.View
-                            style={[styles.toggleDotStyle, toggleImageOnStyle]}
-                        />
-                    </PanGestureHandler>
-                </Animated.View>
+            <Animated.View style={cursorStyle}>
+                <PanGestureHandler onGestureEvent={panGestureHandler}>
```
savelichalex: Shouldn't it be also disabled if needed? 🤔

sertony: Agree

packages/hydrogen/src/UISwitcher/ToggleSwitcher/index.tsx
```diff
@@ -33,46 +27,41 @@ export const ToggleSwitcher: React.FC<UISwitcherProps> = ({
...
+        onPress,
+    );
+
+    const switcherStyle = React.useMemo(() => {
+        return disabled ? styles.disabledSwitcherStyle : toggleBackgroundStyle;
+    }, [disabled, toggleBackgroundStyle]);
 
     return (
-        <RawButton
+        <TapGestureHandler
```
sertony, line 41: Also, shouldn't it `waitFor` the PanGesture handler to avoid processing both Taps and Pans?

ghost: @sertony At first I thought so too, but waitFor={panRef} causes problems on the web(Tap does not fires at all). Without this, the gestures do not conflict.

savelichalex (approved):

sertony (approved):

PR #1268, opened 2021-09-17
Fix cropping top toast notice

SAnatoliiS:


savelichalex (approved):

packages/popups/src/Notice/toastNoticeHooks/useBottomToastNoticeYSnapPoints.ts
```diff
@@ -3,6 +3,9 @@ import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import { UIConstant } from '@tonlabs/uikit.navigation';
 import type { SnapPoints } from '../types';
 
+// @inline
+const INVALID_VALUE = 100000;
```
sertony, line 7: This value is reused across several files. Can we move it at a single spot?

savelichalex: It's `@inline`, so no :)

sertony: Crap :)

sertony (approved):

PR #1272, opened 2021-09-20
Removed legacy UIDropdownAlert

ghost:
https://github.com/tonlabs/TON-Surf/pull/2753


packages/chats/src/UIChatInput/ChatInput.tsx
```diff
@@ -48,11 +47,10 @@ export function ChatInput(props: ChatInputProps) {
         numberOfLinesProp,
         resetInputHeight,
     } = useAutogrowTextView(props.textInputRef, undefined, CHAT_INPUT_NUM_OF_LINES);
-    const showMaxLengthAlert = useChatMaxLengthAlert(MAX_INPUT_LENGTH);
     const { inputHasValue, onChangeText, onKeyPress, onSendText } = useChatInputValue({
         ref: props.textInputRef,
         onSendText: props.onSendText,
-        showMaxLengthAlert,
+        onMaxLength: props.onMaxLength,
```
savelichalex: I'm curious why you decided to move it to the parent, instead of rendering notice here, could you explain your reasoning? 🤔 Seems like an extra step in configuring all that :)

ghost: @savelichalex  Shouldn't we be able to control this Notice? If two or more notices appear in the same position at the same time then we will be able to control their state.

savelichalex: Valid point :) What do you think to have by default it here, and only if a callback is provided give control to a user? 🤔 Just want to reduce things that are need to be configured to use all that

packages/browser/src/UIAddressInput.tsx
```diff
@@ -111,14 +122,33 @@ export function UIAddressInputInternal({
...
+        onMaxLength: onMaxLengthProp || onMaxLength,
     });
     const { validation, onChangeText, clear } = useValidation(
         onBaseChangeText,
         baseClear,
         validateAddress,
     );
 
+    const returnNotice = () => {
+        if (onMaxLengthProp === undefined) {
```
savelichalex:
```suggestion
        if (onMaxLengthProp == null) {
```

packages/browser/src/UIAddressInput.tsx
```diff
@@ -111,14 +122,33 @@ export function UIAddressInputInternal({
...
+                    title={uiLocalized.formatString(
+                        uiLocalized.Chats.Alerts.MessageTooLong,
+                        MAX_INPUT_LENGTH.toString(),
+                    )}
+                    onClose={hideNotice}
+                    duration={UIPopup.Notice.Duration.Long}
+                />
+            );
+        }
+        return;
```
savelichalex:
```suggestion
        return null;
```

packages/browser/src/UIAddressInput.tsx
```diff
@@ -111,14 +122,33 @@ export function UIAddressInputInternal({
...
-        onMaxLength,
+        onMaxLength: onMaxLengthProp || onMaxLength,
     });
     const { validation, onChangeText, clear } = useValidation(
         onBaseChangeText,
         baseClear,
         validateAddress,
     );
 
+    const returnNotice = () => {
```
savelichalex:
```suggestion
    const renderNotice = () => {
```

packages/browser/src/UIAddressInput.tsx
```diff
@@ -100,6 +101,16 @@ export function UIAddressInputInternal({
         resetInputHeight,
     } = useAutogrowTextView(textInputRef, onHeightChange, MAX_INPUT_NUM_OF_LINES);
 
+    const [isNoticeVisible, setNoticeVisible] = React.useState(false);
+
+    const onMaxLength = () => {
```
savelichalex: Let's wrap it with `useCallback` or use it inline instead

ghost: @savelichalex Yes, just for now I'm trying to turn on ESLint... Something is wrong in my VSCode :)

ghost: @savelichalex I will re-request a review when I finish this

savelichalex (approved):

SAnatoliiS (approved):

PR #1275, opened 2021-09-22
Update package.json

BigRoma:
Move react-native-localization and @react-native-clipboard/clipboard to the dependencies section

### DIFF
--- a/packages/localization/package.json
+++ b/packages/localization/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@tonlabs/uikit.localization",
-    "version": "2.3.0",
+    "version": "2.3.1",
     "description": "Localization utilities for Tonlabs's UIKit",
     "scripts": {
         "tsc": "tsc --noEmit",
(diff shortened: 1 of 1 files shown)


SAnatoliiS (approved):

packages/localization/package.json
```diff
@@ -34,13 +34,13 @@
         "src/"
     ],
     "dependencies": {
+        "@react-native-clipboard/clipboard": "^1.8.4",
```
savelichalex, line 37: Why do you need clipboard dependency? I don't see it in code...

packages/localization/package.json
```diff
@@ -34,13 +34,13 @@
...
         "bignumber.js": "9.0.0",
-        "dayjs": "1.9.1"
+        "dayjs": "1.9.1",
+        "react-native-localization": "2.1.6"
```
savelichalex, line 40: The package has native code, it should be placed in peerDeps

PR #1279, opened 2021-09-23
UIAnimatedBalance and UIPushNotice improvements

savelichalex:


packages/flask/src/UIAnimatedBalance/UIAnimatedBalance.tsx
```diff
@@ -120,6 +143,14 @@ export function UIAnimatedBalance({
 
     const formatted = useDerivedValue(() => {
         const integer = Math.floor(valueHolder.value);
+
+        if (maxFractionalDigits <= 0) {
```
savelichalex, line 147: @SAnatoliiS so it turned out, that in Surf it passed as a `number` type, and it was easier to handle such a case, rather than fight with types :D

vetkamak (approved):

ghost (approved):

PR #1285, opened 2021-09-27
New components to draw numbers consistently with proper formatting

savelichalex:


packages/hydrogen/src/Numbers/UIStaticNumber.tsx
```diff
@@ -0,0 +1,81 @@
...
+        );
+    }, [value, decimalAspect, decimalSeparator]);
+
+    const [integerColorStyle, decimalColorStyle] = useNumberStaticStyles(
+        integerColor,
+        decimalColor,
+    );
+
+    return (
+        <Text testID={testID} accessibilityLabel={`${formatted.integer}${formatted.decimal}`}>
```
Khvoya, line 58: Please add '.' between formatted.integer and formatted.decimal

savelichalex: It's [already there](https://github.com/tonlabs/UIKit/pull/1285/files#diff-0c828d2707549d5e6e6129d9104fb10012b0a3625e50ca3e8b5daa9ebbf128a7R113) :)

Khvoya: great, thanks!

ghost (approved):

packages/hydrogen/src/Numbers/UICurrencySign.tsx
```diff
@@ -0,0 +1,168 @@
...
+     * It's placed to have a separation from main balance,
+     * instead of `marginLeft` as space is more "text friendly"
+     * and will layout properly in text
+     */
+    return (
+        <Text style={[Typography[signVariant], styles.iconTextInputContainer]}>
+            {'\u00A0'}
+            <AnimatedUIImage
+                source={signIcon}
+                // resizeMode={'contain'}
```
vetkamak: Is it still needed?

vetkamak (approved):

Example/src/screens/Finances.tsx
```diff
@@ -82,6 +236,15 @@ export function FinancesScreen() {
                 options={{
                     useHeaderLargeTitle: true,
                     title: 'Finances',
+                    headerRightItems: [
+                        Platform.OS === 'ios' && {
+                            label: `${I18nManager.isRTL ? 'Disable' : 'Enable'} RTL`,
+                            onPress: () => {
+                                I18nManager.forceRTL(!I18nManager.isRTL);
+                                NativeModules.DevSettings.reload();
```
sertony, line 253: 👍

packages/hydrogen/src/Numbers/DebugGrid.tsx
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react';
+import { View } from 'react-native';
+import { getFontMesurements, TypographyVariants } from '../Typography';
+
+export function DebugGrid({ variant }: { variant: TypographyVariants }) {
```
sertony, line 5: Interesting stuff :)

packages/hydrogen/src/Numbers/UIAnimatedNumber.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { useTextLikeContainer } from './useTextLikeContainer';
+import { DebugGrid } from './DebugGrid';
+
+Animated.addWhitelistedNativeProps({ text: true });
+
+export function UIAnimatedNumber({
+    testID,
+    children,
+    decimalAspect = UINumberDecimalAspect.None,
+    // appearance
```
sertony:
```suggestion
    // appearance related props >>>
```

packages/hydrogen/src/Numbers/UIAnimatedNumber.tsx
```diff
@@ -0,0 +1,132 @@
...
+        ...Platform.select({
+            web: {},
+            default: { lineHeight: undefined },
+        }),
+    },
+    decimalInput: {
+        fontVariant: ['tabular-nums'],
+        // reset RN styles to have proper vertical alignment
+        padding: 0,
+        ...Platform.select({ web: {}, default: { lineHeight: undefined } }),
```
sertony: I wonder why Prettier formatted `Platform.select` differently here....

packages/hydrogen/src/Numbers/UICurrencySign.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+        /**
+         * Since iOS can't position icon properly on baseline
+         * we do it manually.
+         *
+         * Important to notice here that iOS can't position image
+         * to the bottom, it's actually has some aligment and it
+         * looks like that it's height of descent (see Typography)
+         * so we have to subtract it from baseline
+         */
```
sertony, line 70: Good research

packages/hydrogen/src/Numbers/UICurrencySign.tsx
```diff
@@ -0,0 +1,260 @@
...
+    return (
+        <AnimatedCurrencyIcon
+            loading={loading}
+            signVariant={signVariant}
+            signIcon={signIcon}
+            signIconAspectRatio={signIconAspectRatio}
+            signIconInlineHeight={signIconInlineHeight}
+            signIconAlign={signIconAlign}
+        />
+    );
```
sertony, line 259: What if I change the `loading` prop to `false` while it was loading?

We are to finish the rotation to make it look nice without any glitches (as it was implemented before).

I think this is important. Check out the legacy implementation if needed.

packages/hydrogen/src/Numbers/runOnUILocalizedNumberFormat.ts
```diff
@@ -0,0 +1,148 @@
...
+const DECIMAL_ASPECT_SHORT_ELLIPSIZED = 4;
+
+export enum UINumberDecimalAspect {
+    None = DECIMAL_ASPECT_NONE,
+    Precision = DECIMAL_ASPECT_PRECISION,
+    Short = DECIMAL_ASPECT_SHORT,
+    ShortEllipsized = DECIMAL_ASPECT_SHORT_ELLIPSIZED,
+}
+
+// TODO: move it to uiLocalized!
```
sertony: Is it even possible?

packages/hydrogen/src/Numbers/UIAnimatedNumber.tsx
```diff
@@ -0,0 +1,132 @@
...
+    // appearance
+    integerVariant,
+    integerColor,
+    decimalVariant,
+    decimalColor,
+    sign,
+    showDebugGrid,
+    showPositiveSign,
+}: UINumberGeneralProps & UINumberAppearance & { sign?: React.ReactNode }) {
+    const value: number = useNumberValue(children, decimalAspect);
```
sertony: I'm scared with the fact that we are to convert BigNumbers to Numbers here, since we can easily loose the precision for nano tokens

I suggest an idea to parse the BigNumber and get its `integer` and `decimal` parts as Numbers (they must fit `MAX_SAFE_INTEGER`) and hence we'll be able to avoid any issues.

```suggestion
    const { integer, decimal } = useNumberValue(children, decimalAspect);
```

packages/hydrogen/src/Numbers/runOnUILocalizedNumberFormat.ts
```diff
@@ -0,0 +1,148 @@
...
+     * We take it with a highest precision possible to not lose data
+     * then on formatting we're still going to apply correct one
+     */
+    const decimalDigits = getDecimalAspectDigits(DECIMAL_ASPECT_PRECISION);
+
+    return value
+        .multipliedBy(10 ** decimalDigits)
+        .integerValue()
+        .dividedBy(10 ** decimalDigits)
+        .toNumber();
```
sertony: Not sure if I understand the difference between this and the following one
```suggestion
    return value.toFixed(decimalDigits).toNumber();
```

packages/hydrogen/src/Numbers/UIAnimatedNumber.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+    React.useEffect(() => {
+        if (valueHolder.value === value) {
+            return;
+        }
+
+        valueHolder.value = withTiming(value, {
+            duration: 400,
+            easing: Easing.inOut(Easing.ease),
+        });
```
sertony: Too bad it doesn't work with BigNumber =/

But instead you can animate "integer" and "decimal" parts independently well.

packages/hydrogen/src/Numbers/useNumberValue.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+import BigNumber from 'bignumber.js';
+import { bigNumToNumber, UINumberDecimalAspect } from './runOnUILocalizedNumberFormat';
+
+export function useNumberValue(rawValue: number | BigNumber, decimalAspect: UINumberDecimalAspect) {
+    return React.useMemo(
+        () => (rawValue instanceof BigNumber ? bigNumToNumber(rawValue, decimalAspect) : rawValue),
+        [rawValue],
+    );
```
sertony: Please, think of changes suggested above (i.e. to split the rawValue into two parts and operate them safely)

packages/localization/src/service.ts
```diff
@@ -61,7 +62,12 @@ export class LocalizationService<T> extends (LocalizedStringsService as Localize
     languages: Language[];
     localeInfo: StringLocaleInfo;
 
-    constructor({ languages, localeInfo = defaultLocaleInfo }: LanguageServiceOptions<T>) {
+    constructor({
+        languages,
+        localeInfo = NativeModules.UIKitLocalization != null
+            ? NativeModules.UIKitLocalization.getConstants()
```
sertony, line 68: How about creating an independent and exportable method for this?

Since we might also want to use it in Surf (once you remove the same native code from that side in this PR: https://github.com/tonlabs/TON-Surf/pull/2778/files)

sertony (changes requested):

packages/hydrogen/src/Numbers/UIStaticNumber.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    const formatted = React.useMemo(() => {
+        return runOnUILocalizedNumberFormat(
+            value,
+            decimalAspect,
+            decimalSeparator,
+            integerGroupChar,
+            showPositiveSign,
+        );
+    }, [value, decimalAspect, decimalSeparator, integerGroupChar, showPositiveSign]);
```
sertony: How about dealing with `BigNumber` here only, since we don't need to animate it using `withTiming`..

packages/localization/src/service.ts
savelichalex: I actually think that it might be a good idea not to expose it, and use `localeInfo` directly :)

packages/hydrogen/src/Numbers/UIAnimatedNumber.tsx
```diff
@@ -37,31 +35,94 @@ export function UIAnimatedNumber({
...
-            decimalSeparator,
-            integerGroupChar,
-            showPositiveSign,
+        diff.current = value.minus(prevValueHolder.current);
+        progress.value = 0;
+        progress.value = withTiming(
+            1,
+            {
+                duration: 400,
+                // easing: Easing.inOut(Easing.ease),
```
sertony: Is it glitchy with it and the callback?

packages/localization/src/service.ts
sertony: But we use it pretty often in Surf (usually we call `TONString.getLocaleInfo()`)

To do so we have to setup our TONUtility with the proper localeInfo received from native

savelichalex:
> But we use it pretty often in Surf (usually we call TONString.getLocaleInfo())
>
> To do so we have to setup our TONUtility with the proper localeInfo received from native

@sertony I guess we should start using one thing for that, and I personally vote for `uiLocalization` :)

sertony:
> > But we use it pretty often in Surf (usually we call TONString.getLocaleInfo())
> > To do so we have to setup our TONUtility with the proper localeInfo received from native
> 
> @sertony I guess we should start using one thing for that, and I personally vote for `uiLocalization` :)

Me too, but without that feature it will become breaking (once we remove the duplicated native code from Surf)

savelichalex:
> > > But we use it pretty often in Surf (usually we call TONString.getLocaleInfo())
> > > To do so we have to setup our TONUtility with the proper localeInfo received from native
> > 
> > 
> > @sertony I guess we should start using one thing for that, and I personally vote for `uiLocalization` :)
> 
> Me too, but without that feature it will become breaking (once we remove the duplicated native code from Surf)

I'm not quite understand what will be breaking 😕

packages/hydrogen/src/Numbers/UIAnimatedNumber.tsx
savelichalex: Nope, just a thing that I forgot to return when experimented with `withTiming` and `BigNumber`

packages/localization/src/service.ts
savelichalex: I mean it's public, you can easily use `uiLocalization.localeInfo` to access it

packages/hydrogen/src/Numbers/UICurrencySign.tsx
savelichalex: Yep, done

packages/localization/src/service.ts
sertony: Got it. OK, we'll switch on it.

sertony (approved):

PR #1295, opened 2021-09-30
Fix Scrollable for react-native-pager-view

ghost:
In a react-native-pager-view(UIPagerView, UICarouselView) the pan gesture was intercepted by a scroll on android


savelichalex:
Unfortunately we can't merge it right now, as it would break LargeHeader, I have to think how to do it better

PR #1296, opened 2021-09-30
Fix pager ripple effect

ghost:


savelichalex:
So, what is the fix exactly? :)

packages/navigation/src/UIPagerView/UIPagerViewContainer.tsx
```diff
@@ -337,18 +339,15 @@ const styles = StyleSheet.create({
...
     },
     fixedTabBar: {
-        height: UIConstant.pagerViewHeight,
+        height: UIConstant.pager.pagerViewHeight,
         backgroundColor: 'transparent',
         shadowColor: 'transparent',
         elevation: 0,
         justifyContent: 'center',
+        bottom: UIConstant.pager.tabBarOffset,
+        borderWidth: 0,
```
ghost, line 355: @savelichalex This is the core of the fix. The rest of the styles are just to improve the look.

savelichalex: Did you manage to understand how it works? Because I don't quite understand how it fixes the problem :)

ghost: @savelichalex This is just a hack, I'm not sure why it works at all :)

savelichalex: Please leave a comment then with breadcrumbs with 1. what was the problem 2. how to reproduce it 3. how it fixes it (here you can write that there is some magic behind)

packages/navigation/src/UIPagerView/UIPagerViewContainer.tsx
```diff
@@ -345,6 +345,11 @@ const styles = StyleSheet.create({
         elevation: 0,
         justifyContent: 'center',
         bottom: UIConstant.pager.tabBarOffset,
+        /**
+         * The ripple effect on android fills all the available space.
+         * This is visible if the page has a transparent background.
+         * borderWidth: 0 - brings some kind of magic in the form of limiting the effect to invisible borders
```
savelichalex:
```suggestion
         * The ripple effect on android fills all the available space.
         * But there was a problem when ripple effect went outside of borders,
         * it is visible if the page has a transparent background.
         *
         * borderWidth: 0 - brings some kind of magic in the form of limiting the effect to visible borders
```

savelichalex (approved):

vetkamak (approved):

PR #1300, opened 2021-10-04
UIGridView

ghost:


packages/navigation/src/UIGridView/index.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { ListRenderItem, View, StyleSheet } from 'react-native';
+import { FlatList } from '..';
+import { UIConstant } from '../constants';
+import type { GridProps } from './types';
+
+export function UIGridView<T>({
```
savelichalex: What do you think to call it `UIGridList`? :)

packages/navigation/src/UIGridView/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+    itemHeight,
+    testID,
+}: GridProps<T>) {
+    const [squareSize, setSquareSize] = React.useState({ width: 0, height: 0 });
+
+    const renderItem: ListRenderItem<T> = React.useCallback(
+        ({ item, index, separators }) => {
+            return (
+                <View style={[styles.item, squareSize]}>
+                    {renderItemProp?.({ item, index, separators })}
```
savelichalex: I guess there is no need to make optional check since the prop is not optional

packages/navigation/src/UIGridView/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+                </View>
+            );
+        },
+        [renderItemProp, squareSize],
+    );
+
+    const onLayout = React.useCallback(
+        (event: any) => {
+            const { width } = event.nativeEvent.layout;
+            const squareDimension = width / UIConstant.grid.numColumns - 24;
```
savelichalex: Let's move constant, btw why it's not a `contentOffset`? Looks inconsistent

packages/navigation/src/UIGridView/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+            setSquareSize({
+                width: squareDimension,
+                height: itemHeight ?? squareDimension,
+            });
+        },
+        [itemHeight],
+    );
+
+    return (
+        <View testID={testID} style={[styles.container, containerStyle]} onLayout={onLayout}>
```
savelichalex: I'm not sure that we should wrap it with View, first `padding` is going to break scrollingIndicator from a right edge, second `flex: 1` makes impossible to have intrinsic size. I would suggest to remove it and use plain FlatList

packages/navigation/src/UIGridView/index.tsx
ghost: It was optional

packages/navigation/src/UIGridList/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+                <View style={[styles.item, height]}>
+                    {renderItemProp({ item, index, separators })}
+                </View>
+            );
+        },
+        [renderItemProp, itemHeight],
+    );
+
+    return (
+        <FlatList
```
savelichalex, line 39: I also thought, that since we know height, then we can apply optimisation by default! I'm talking about https://reactnative.dev/docs/0.64/flatlist#getitemlayout

ghost: @savelichalex but we do not know the height if it is not passed by itemHeight. By default used aspectRatio: 1 and flex: 0.5. You would like to calculate the height if passed itemHeight?

savelichalex: Maybe apply it only if there was height provided? 🤔

savelichalex (approved):

vetkamak (approved):

vetkamak:
@wearedusa let's merge it to surf@6.4.4, @savelichalex do you have any objections?

PR #1305, opened 2021-10-06
Parsed url improvements

ghost:


packages/chats/src/constants.ts
```diff
@@ -1,5 +1,7 @@
 import { Platform } from 'react-native';
 
+export const URI_TON_SURF = 'uri.ton.surf';
```
sertony: I definitely don't like we have it in constants. This should be removed and passed via React Context instead, IMHO.

packages/chats/src/BubblePlainText.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable react/destructuring-assignment */
+/* eslint-disable react/jsx-props-no-spreading */
```
sertony: This is not required anymore as we disabled it globally ;)

packages/chats/src/BubblePlainText.tsx
```diff
@@ -26,6 +28,29 @@ import { MessageStatus, OnLongPressText, OnPressUrl } from './types';
...
+        if (domain === URI_TON_SURF) {
+            const clearUrl = matches[0].replace(matches[1], '');
+            return matchingString.replace(matches[0], '') + decodeURI(clearUrl);
+        }
+    }
+    return matchingString;
+};
+
+const formatURI = (str: string) => {
+    const encodedURIArray = encodeURI(str).match(PATTERNS.url);
```
sertony: What if the text is not just an URL, but some text + URL, then in this case `encodeURI` should fail AFAIR.

packages/chats/src/BubblePlainText.tsx
```diff
@@ -26,6 +28,29 @@ import { MessageStatus, OnLongPressText, OnPressUrl } from './types';
...
+    if (encodedURIArray) {
+        const encodedURI = encodedURIArray[0];
```
sertony: What if there are more than one URI?

packages/chats/src/BubblePlainText.tsx
```diff
@@ -26,6 +28,29 @@ import { MessageStatus, OnLongPressText, OnPressUrl } from './types';
...
+            const clearUrl = matches[0].replace(matches[1], '');
+            return matchingString.replace(matches[0], '') + decodeURI(clearUrl);
+        }
+    }
+    return matchingString;
+};
+
+const formatURI = (str: string) => {
+    const encodedURIArray = encodeURI(str).match(PATTERNS.url);
+    if (encodedURIArray) {
```
sertony:
```suggestion
    if (encodedURIArray && encodedURIArray.length) {
```

packages/chats/src/BubblePlainText.tsx
```diff
@@ -26,6 +28,29 @@ import { MessageStatus, OnLongPressText, OnPressUrl } from './types';
...
+        const encodedURI = encodedURIArray[0];
+        const clearText = str.replace(decodeURI(encodedURI), '');
+        return clearText + encodedURI;
```
sertony: Can you add a comment that here we tend to reshuffle the original text by putting the string at the end of the message?

packages/chats/src/BubblePlainText.tsx
```diff
@@ -26,6 +28,29 @@ import { MessageStatus, OnLongPressText, OnPressUrl } from './types';
 import type { ChatPlainTextMessage, PlainTextMessage } from './types';
 import { useBubblePosition, useBubbleContainerStyle } from './useBubblePosition';
 import { useBubbleBackgroundColor, useBubbleRoundedCornerStyle } from './useBubbleStyle';
+import { URI_TON_SURF } from './constants';
+
+const renderParsedText = (matchingString: string, matches: string[]) => {
+    if (matches[1]) {
+        const domain = matches[0].replace(matches[1], '').replace(matches[2], '');
```
sertony: Something is apparently wrong here..
First, matches[2] might be undefined
Second, matches refers not just to URLs, but it can be used with phones numbers, email, etc as per https://github.com/taskrabbit/react-native-parsed-text documentation

@savelichalex please, take a look

packages/chats/src/BubblePlainText.tsx
```diff
@@ -26,6 +28,29 @@ import { MessageStatus, OnLongPressText, OnPressUrl } from './types';
...
+        if (domain === URI_TON_SURF) {
+            const clearUrl = matches[0].replace(matches[1], '');
+            return matchingString.replace(matches[0], '') + decodeURI(clearUrl);
+        }
```
sertony: To be honest, it's not clear to me what's going on here. As far as I see we again tend to display the URI at the end of the message, am I right?

sertony (changes requested):

packages/chats/src/BubblePlainText.tsx
savelichalex: It mustn't be that, that's for sure. User won't expect `url` to be put at the end of a message

packages/chats/src/BubblePlainText.tsx
ghost: @sertony No, I tested this case

packages/chats/src/BubblePlainText.tsx
wearedusa: Matches come here one at a time

packages/chats/src/BubblePlainText.tsx
ghost: @sertony The match comes from the selected pattern. In this case, url. But I already rewrote it more simply, without matches[]

packages/chats/src/constants.ts
ghost: @sertony  If so then maybe it should be an array of domains? In case you want to transfer several "friendly domains" 🤔

sertony: Yep, I think so

ghost: @sertony @savelichalex  Done. I also merged onUrlPress and safeURLs into one context for more consistency.

packages/chats/src/constants.ts
```diff
@@ -2,6 +2,12 @@ import { Platform } from 'react-native';
 
 const GRID_LINE_HEIGHT = 4;
 
+export const RegExpConstants = {
+    domain: /.+\/\/|www.|\/.+/g,
+    protocol: /^https?:\/\//,
+    url: /(https?:\/\/|www\.)[-а-яА-Яa-zA-Z0-9@:%._\+~#=]{1,256}\.[a-z]{2,6}\b([-а-яА-Яa-zA-Z0-9@:%_\+.~#?&\/=]*[-а-яА-Яa-zA-Z0-9@:%_\+~#?&\/=])*/i,
```
savelichalex, line 8: So, it's gonna work only with english and russian?

savelichalex: What about other languages?

savelichalex: I thought that you wanted to use `encodeURI` for escaping the link. Maybe we can put it in Input, when user paste link there

Example/src/screens/Browser.tsx
```diff
@@ -123,12 +123,15 @@ const BrowserScreen = React.forwardRef<BrowserScreenRef>((_props, ref) => {
 
     const { height } = useWindowDimensions();
 
+    const safeURLsArray = ['uri.ton.surf'];
```
savelichalex, line 126: Looks like with modified regexp, this is not needed anymore, is that right?

ghost: @savelichalex No. We have to pass a list of safe domains. Regexp checks that the link is a link. And then we check that the domain belongs to any of the safe domains passed to safeURLsArray

savelichalex: I don't think that we should have this logic actually, as it doesn't matter for us, if it's safe or not, we should just parse any link

packages/chats/src/constants.ts
ghost: @savelichalex 
>What about other languages?

I thought the task was to add the Russian language :)

Example/src/screens/Browser.tsx
ghost: @savelichalex  Safe links don't work the way you think. A safe link means that we can choose not to show the protocol for it. We show all links as links. Just safe links without protocol. That was the task.

packages/chats/src/constants.ts
ghost: If we want to parse links with all languages, then I think a separate task is needed for this.

Example/src/screens/Browser.tsx
savelichalex: The task was to try to parse ALL links without protocol, and only if it isn't possible use specific domains

ghost: @savelichalex Oh. Okay. Well then it was worth describing the task in more detail. Because I memorized it orally as "We need our domain to be without a protocol"

savelichalex: quote from the task:
> please also take a look how to parse links without http(s) properly

ghost: @savelichalex ok i looked

PR #1306, opened 2021-10-06
Fix grid last item

ghost:


savelichalex:
What the fix doing? How exactly it fixes the problem?

packages/navigation/src/UIGridList/index.tsx
```diff
@@ -54,6 +55,10 @@ const styles = StyleSheet.create({
         margin: UIConstant.contentOffset / 2,
         flex: 0.5,
     },
+    lastItem: {
```
savelichalex: What if last item is actually second in a row? will it work?

ghost: Oh and you're right

vetkamak (approved):

savelichalex (approved):

PR #1307, opened 2021-10-06
UIMasonryList

savelichalex:


packages/navigation/src/index.ts
```diff
@@ -1,6 +1,8 @@
 import './useAutoFocus';
 
 export * from './UIGridList';
+// TODO: move it
```
sertony: Will you?)

packages/navigation/src/UIGridList/UIMasonryList.tsx
```diff
@@ -0,0 +1,498 @@
...
+function measureList<Item>(cellWidth: number, data: MasonryItem<Item>[], numOfColumns: number) {
+    const columnsHeights = new Array(numOfColumns).fill(0);
+
+    let maxHeight = 0;
+
+    const layout: Layout = {};
+
+    // Calculating offsets for columns in advance,
+    // since for each row they are simply going to
+    // be repeated agan
```
sertony:
```suggestion
    // be repeated again
```

packages/navigation/src/index.ts
savelichalex: I actually don't know, what do you think, should I? Or it's good enough to keep it there for now? :)

packages/navigation/src/UIGridList/UIMasonryList.tsx
```diff
@@ -0,0 +1,498 @@
...
+
+        columnsHeights[column] += height + interColumnOffset;
+
+        // eslint-disable-next-line no-param-reassign
+        layout[item.key] = {
+            height,
+            x,
+            y,
+        };
+    }
```
sertony: Can you add a comment that this layout algo might change in the future as it has some cons as having one column way longer then another (i.e. when having several "long" images in one column)?

packages/navigation/src/UIGridList/UIMasonryList.tsx
```diff
@@ -0,0 +1,498 @@
...
+            right = middle;
+        } else {
+            left = middle;
+        }
+
+        if (right - left <= 1) {
+            if (indexes[left].y >= coord) {
+                return left;
+            }
+            return right;
```
sertony: 👍

packages/navigation/src/UIGridList/UIMasonryList.tsx
```diff
@@ -0,0 +1,498 @@
...
+
+type MasonryCellsRefs = Record<string, React.RefObject<UIMasonryListCellRef>>;
+
+function getMasonryCellsRefs<Item>(data: MasonryItem<Item>[]) {
+    return data.reduce<MasonryCellsRefs>((acc, { key }) => {
+        if (acc[key] != null) {
+            return acc;
+        }
+        acc[key] = React.createRef<UIMasonryListCellRef>();
+        return acc;
```
sertony: Not sure if there is any benefit, just a shorter record (maybe it's simpler to read though).
```suggestion
        if (acc[key] == null) {
            acc[key] = React.createRef<UIMasonryListCellRef>();
        }
        return acc;
```

packages/navigation/src/UIGridList/UIMasonryList.tsx
```diff
@@ -0,0 +1,498 @@
...
+    }
+
+    React.useEffect(() => {
+        cellsRefs.current = getMasonryCellsRefs(data);
+    }, [data]);
+
+    return cellsRefs.current;
+}
+
+const virtualWindowSize = 1.5;
```
sertony: Can you move it somewhere on top and add some comment for people to understand why do we need it for when doing virtualization?

packages/navigation/src/UIGridList/UIMasonryList.tsx
```diff
@@ -0,0 +1,498 @@
...
+                },
+            } = event;
+            /**
+             * This is a core of virtualization.
+             *
+             * At the point we already prepared a sorted index
+             * which we can easily and fast traverse to find
+             * cells keys that fit into our virtual window.
+             *
+             * We use modified binary search alrorithm,
```
sertony:
```suggestion
             * We use modified binary search algorithm,
```

packages/navigation/src/UIGridList/UIMasonryList.tsx
```diff
@@ -0,0 +1,498 @@
...
+             * next measurement.
+             * So, we have
+             * - previous top item
+             * - current top item
+             * - previous bottom item
+             * - current bottom item
+             *
+             * Since our index is sorted, we can easily find out that
+             * items between previous and current items the ones that're
+             * changed, and others shouldn't be touched.
```
sertony: 🔥

packages/navigation/src/UIGridList/UIMasonryList.tsx
```diff
@@ -0,0 +1,498 @@
...
+                    item={item}
+                    renderItem={renderItem}
+                    width={cellWidth}
+                    layout={layout[item.key]}
+                />
+            ));
+
+        return (
+            <ScrollView
+                // eslint-disable-next-line react/jsx-props-no-spreading
```
sertony: This one is not needed anymore ;)

sertony (approved):
Great job! The component should be great in terms of optimization!

ghost (approved):

packages/navigation/src/index.ts
sertony: Maybe it's better to move both in some kind of Lists folder?

sertony (approved):

PR #1308, opened 2021-10-07
fix formatting

sertony:


packages/components/package.json
```diff
@@ -35,6 +35,8 @@
     "dependencies": {
         "@tonlabs/uikit.assets": "^2.3.0",
         "@tonlabs/uikit.core": "^2.3.0",
+        "@tonlabs/uikit.hydrogen": "^2.3.0",
+        "@tonlabs/uikit.localization": "^2.3.0",
```
sertony, line 39: This was broken

packages/flask/package.json
```diff
@@ -39,6 +39,7 @@
         "@tonlabs/uikit.hydrogen": "^2.3.0",
         "@tonlabs/uikit.localization": "^2.3.0",
         "@tonlabs/uikit.navigation": "^2.3.0",
+        "dayjs": "1.9.1",
```
sertony, line 42: This was broken too

packages/flask/package.json
```diff
@@ -66,6 +67,7 @@
         "react-native": "^0.65.0",
         "react-native-gesture-handler": "^1.10.3",
         "react-native-reanimated": "^2.2.1",
+        "react-native-safe-area-context": "^3.1.3",
```
sertony, line 72: And this

packages/navigation/src/Sheets/UISheet/usePosition.tsx
```diff
@@ -77,7 +77,7 @@ export function usePosition(
     const snapPointPosition = useDerivedValue(() => {
         let snapPoint = 0 - height.value;
         if (shouldHandleKeyboard) {
-            snapPoint - keyboardHeight.value;
+            snapPoint -= keyboardHeight.value;
```
sertony: Oops

Example/__tests__/App-test.js
```diff
@@ -4,10 +4,11 @@
 
 import 'react-native';
 import React from 'react';
+import renderer from 'react-test-renderer';
+
 import App from '../App';
 
 // Note: test renderer must be required after react-native.
-import renderer from 'react-test-renderer';
```
savelichalex, line 10: Probably comment also should be moved :)

packages/core/src/UITextStyle/UIStyleText.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import UITextStyle from '../UITextStyle';
+import UITextStyle from '.';
```
savelichalex, line 2: hm, this might introduce cycles 🤔

packages/flask/src/UICountryPicker/ListEmptyComponent.tsx
```diff
@@ -24,6 +25,7 @@ export const ListEmptyComponent = () => {
     const { loading } = React.useContext(CountryPickerContext);
 
     const renderContent = React.useMemo(
+        // eslint-disable-next-line no-confusing-arrow
         () => (loading ? renderLoading() : renderEmptyList()),
```
savelichalex:
```suggestion
        () => {
            if (loading) {
                return renderLoading();
            }
            return renderEmptyList();
        },
```

packages/hydrogen/src/UIImage/UIImage.web.tsx
```diff
@@ -16,8 +16,8 @@ import type { UIImageProps } from './types';
 
 const useImageDimensions = (style: StyleProp<ImageStyle>, source: any) => {
     return React.useMemo(() => {
-        let width: number | string | undefined = undefined;
-        let height: number | string | undefined = undefined;
+        let width: number | string | undefined;
```
savelichalex, line 19: Actually TS usually complain if it hadn't initial value 🤔

packages/hydrogen/src/UIMsgButton.tsx
```diff
@@ -379,7 +379,7 @@ export const UIMsgButton = ({
                     {iconPosition === UIMsgButtonIconPosition.Left && icon != null && (
                         <Button.Icon
                             source={icon}
-                            style={!!title ? styles.leftIcon : null}
+                            style={title ? styles.leftIcon : null}
```
savelichalex, line 382: Maybe there was a case for the check? 🤔

packages/popups/src/Notice/toastNoticeHooks/useTopToastNoticeYSnapPoints.ts
```diff
@@ -19,10 +19,10 @@ export const useTopToastNoticeYSnapPoints = (
         /**
          * Let's be safe if initialWindowMetrics does not has the value
          */
-        return !!initialWindowMetrics?.frame.height
+        return initialWindowMetrics?.frame.height
```
savelichalex:
```suggestion
        return initialWindowMetrics?.frame.height != null
```

packages/core/src/UITextStyle/UIStyleText.js
sertony: Actually it doesn't due to the fact that UIStyleText is not imported in index.
I don't want to make it nice since it's a legacy code :)

packages/hydrogen/src/UIImage/UIImage.web.tsx
sertony: It's OK since not having any initial value means `undefined` which is here by type annotation

packages/hydrogen/src/UIMsgButton.tsx
sertony: It's OK since it's not `!!title && styles.leftIcon` :)

packages/hydrogen/src/UIImage/UIImage.web.tsx
savelichalex: I know, but TS compiler sometimes complain about it, not me :D

packages/popups/src/Notice/toastNoticeHooks/useTopToastNoticeYSnapPoints.ts
sertony: Actually in becomes different now. 
If the height is zero it will use zero height instead of `screenDimensionsHeight` as it was before. Is it what you want?

sertony: Anyway I've applied the changes as you suggested, as they seem to be justified

packages/hydrogen/src/UIImage/UIImage.web.tsx
sertony: Well, it didn't, I've checked)

packages/popups/src/Notice/toastNoticeHooks/useTopToastNoticeYSnapPoints.ts
savelichalex: I would then put `> 0`, if it was the case :)

savelichalex (approved):

ghost (approved):

PR #1311, opened 2021-10-07
test ids for browser images

Khvoya:


packages/browser/src/MediaOutput.tsx
```diff
@@ -47,7 +47,7 @@ export const MediaOutput = ({
         }
     }, [onOutput]);
     return (
-        <View onLayout={onLayout}>
+        <View testID={`media_${prompt || 'default'}`} onLayout={onLayout}>
```
savelichalex: Do we really need prompt here? It's optional, and you still might be in a situation, when there will be few elements with the same `testID`. Besides, it would be hard to follow whether someone changed it or not

Khvoya: Suggest something else as a unique identifier

savelichalex: Maybe ask debot to provide it? Only debot can provide unique.
Or we can use image url.

savelichalex (approved):

packages/chats/src/Lightbox/DuplicateContent.tsx
```diff
@@ -169,7 +169,10 @@ export const DuplicateContent = ({
                             <View onLayout={onLayout}>{previewImage}</View>
                         </Animated.View>
                         {isFullSizeDisplayed ? (
-                            <Animated.View style={styles.fullSizeImage}>
+                            <Animated.View
+                                testID={`image_expanded`}
```
sertony: You ESLint seem not to work properly in `surf@6.4.4` ;)
```suggestion
                                testID="image_expanded"
```

sertony (approved):

PR #1320, opened 2021-10-18
Fixed day size

ghost:
Safari does not support aspect-ratio prior to Safari 15
https://caniuse.com/?search=aspect-ratio
Fix for this: 
https://ibb.co/8BTjd65

### DIFF
--- a/packages/flask/src/UIDateTimePicker/Calendar/Calendar.tsx
+++ b/packages/flask/src/UIDateTimePicker/Calendar/Calendar.tsx
@@ -9,6 +9,8 @@ import {
     useTheme,
     TouchableOpacity,
     PortalManager,
+    getFontMesurements,
+    TypographyVariants,
 } from '@tonlabs/uikit.hydrogen';
 import { uiLocalized } from '@tonlabs/uikit.localization';
 
@@ -51,6 +53,19 @@ const Column = React.memo(function Column({
     onSelect: (column: number, row: number) => void;
 }) {
     const theme = useTheme();
+
+    /**
+     * Safari below 15 version not support aspect-ratio. So we have to provide width/height for day view.
+     * To create a smooth circle, we need to get the font height.
+     * getFontMesurements() may return undefined, so we must also pass the default value.
+     * The size is calculated for two numbers, so this value needs to be multiplied by two.
+     */
+    const fontHeight =
+        getFontMesurements(TypographyVariants.Action)?.upperline ||
+        UIConstant.calendar.dayCellDefaultFontHeight;
+    const daySize = fontHeight * 2;
+
+    console.log(daySize);
     return (
         <View style={styles.column}>
             {items.map((day, index) => {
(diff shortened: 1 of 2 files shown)


packages/flask/src/constants/index.ts
```diff
@@ -63,6 +63,7 @@ export const UIConstant = {
     calendar: {
         dayCellPadding: 5,
         dayCellPaddingBorderRadius: 20,
+        daySize: 34,
```
savelichalex: I actually didn't want to use the exact size, because
1. it makes impossible to use dynamic font size (i.e. for accessibility)
2. it might break on every font size change

Do you think we can come up with a better idea, how to achieve it without providing specific size? If it's impossible then it's ok, but I really want to think twice about it before merging

ghost: @savelichalex 
1. write a function to calculate font height? 🤔 May be might even work if wrap the view of day one time and get the height from onLayout to avoid using complex functions calculating font size
2. Use https://github.com/airamrguez/react-native-measure-text

savelichalex: we actually already have this :) https://github.com/tonlabs/UIKit/blob/development/packages/hydrogen/src/Typography.ts#L262

packages/flask/src/UIDateTimePicker/Calendar/Calendar.tsx
```diff
@@ -68,8 +83,14 @@ const Column = React.memo(function Column({
 
                 if (day.type === 'dayFiller') {
                     return (
-                        // eslint-disable-next-line react/no-array-index-key
-                        <View style={styles.day} key={`${selectedRow}${index}`}>
+                        <View
+                            style={[
+                                styles.day,
+                                { width: daySize, height: daySize, borderRadius: daySize },
```
savelichalex: I'm not sure that `dayFiller` must have `borderRadius` as it's invisible anyway 🤔

packages/flask/src/UIDateTimePicker/Calendar/Calendar.tsx
```diff
@@ -51,6 +53,19 @@ const Column = React.memo(function Column({
...
+    /**
+     * Safari below 15 version not support aspect-ratio. So we have to provide width/height for day view.
+     * To create a smooth circle, we need to get the font height.
+     * getFontMesurements() may return undefined, so we must also pass the default value.
+     * The size is calculated for two numbers, so this value needs to be multiplied by two.
+     */
+    const fontHeight =
+        getFontMesurements(TypographyVariants.Action)?.upperline ||
+        UIConstant.calendar.dayCellDefaultFontHeight;
+    const daySize = fontHeight * 2;
```
savelichalex:
```suggestion
    const style = React.useMemo(() => {
        const fontHeight = getFontMesurements(TypographyVariants.Action)?.upperline || UIConstant.calendar.dayCellDefaultFontHeight;
        const daySize = fontHeight * 2;
        return { width: daySize, height: daySize };
    }, []);
```
Btw, why are you multiplying `fontHeight` by 2, isn't it should be just adding padding multiplied by 2 (i.e. `daySize = fontHeight + (padding * 2)`)?

ghost: @savelichalex yeah you're right better to use padding*2

packages/flask/src/UIDateTimePicker/Calendar/Calendar.tsx
```diff
@@ -51,6 +53,22 @@ const Column = React.memo(function Column({
     onSelect: (column: number, row: number) => void;
 }) {
     const theme = useTheme();
+
+    /**
+     * Safari below 15 version not support aspect-ratio. So we have to provide width/height for day view.
+     * To create a smooth circle, we need to get the font height.
+     * getFontMesurements() may return undefined, so we must also pass the default value.
+     * Also we have to provide padding size multiplied by two.
```
savelichalex:
```suggestion
     * Also we have to provide padding size multiplied by two, according to our design.
```

savelichalex (approved):

sertony (approved):

sertony:
Changed the base to `surf@6.5.0` as we are fixing the bugs in Surf for release version ;)

PR #1322, opened 2021-10-18
Re-mount an icon in UICurrency when it's changed

savelichalex:


packages/hydrogen/src/Numbers/UICurrencySign.tsx
```diff
@@ -235,12 +244,15 @@ export function UICurrencySign({
         return null;
     }
 
+    const assetSource = resolveAssetSource(signIcon);
```
sertony: How about `useMemo`?

sertony (approved):

SAnatoliiS (approved):

PR #1324, opened 2021-10-19
test id changed

Khvoya:


packages/hydrogen/src/Numbers/UIAnimatedNumber.tsx
```diff
@@ -152,12 +152,9 @@ export function UIAnimatedNumber({
     const textLikeContainer = useTextLikeContainer();
 
     return (
-        <View
-            style={textLikeContainer}
-            testID={testID}
-            accessibilityLabel={`${formatted.value.integer}${formatted.value.decimal}`}
```
savelichalex, line 158: Are you sure we should drop it? Maybe it make sense to leave it for accessibility?

sertony: Agree, probably this was removed accidentally

Khvoya: It contains a wrong value. On the main screen with a positive balance accessibility label is '0.00' (should be ~19k) and I decided to drop it

sertony: @savelichalex do you understand why it was 0.00? Cause it's strange

savelichalex: Yep, it isn't an animated prop, I guess that's why

savelichalex: Anyway, let's not drop it, better to put TODO comment that it has incorrect value, or otherwise we will forget that we have to deal with it

Khvoya: done

sertony (approved):

PR #1326, opened 2021-10-20
Fix: Touchable hitSlop from RNGH does not work on android. 

ghost:
Touchable hitSlop from RNGH does not work on android. Сhanged to a TouchableOpacity from the RN.
Do we need to deal with this or should we wait for the fix?
https://github.com/software-mansion/react-native-gesture-handler/issues/1242

### DIFF
--- a/packages/navigation/src/UIHeaderItems.tsx
+++ b/packages/navigation/src/UIHeaderItems.tsx
@@ -113,6 +113,11 @@ function UIHeaderItemPressable({
     return (
         <>
             <View style={applyMargin ? styles.headerItemMargin : null} />
+            {/**
+             * We are should use TouchableOpacity from RN instead of @tonlabs/uikit.hydrogen
+             * Because  of bug in RNGH:
+             * https://github.com/software-mansion/react-native-gesture-handler/issues/1242
+             */}
             <TouchableOpacity
                 testID={testID}
                 hitSlop={{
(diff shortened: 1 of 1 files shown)


savelichalex:
Please leave a comment with a link to the issue, and please create a task to investigate the issue further, as we definitely have cases when we must use ones from RNGH

ghost:
@savelichalex comment added

packages/navigation/src/UIHeaderItems.tsx
```diff
@@ -113,6 +113,11 @@ function UIHeaderItemPressable({
     return (
         <>
             <View style={applyMargin ? styles.headerItemMargin : null} />
+            {/**
+             * We are should use TouchableOpacity from RN instead of @tonlabs/uikit.hydrogen
+             * Because  of bug in RNGH:
+             * https://github.com/software-mansion/react-native-gesture-handler/issues/1242
```
savelichalex:
```suggestion
             * We are using TouchableOpacity from RN instead of @tonlabs/uikit.hydrogen
             * Because of a bug in RNGH
             * https://github.com/software-mansion/react-native-gesture-handler/issues/1242
             * but we need to switch it back as soon as we resolve it!
```

savelichalex (approved):

sertony (approved):

PR #1331, opened 2021-10-26
Add onEndReached prop for  UIMasonryList

ghost:


kit/scrolls/src/Lists/UIMasonryList.tsx
```diff
@@ -335,6 +339,28 @@ function useVirtualization<Item>(
         [initialWindowHeight, onLayoutProp, setWidth],
     );
 
+    const maybeCallOnEndReached = React.useCallback(
+        (contentLength: number, visibleLength: number) => {
+            const distanceFromEnd = contentLength - visibleLength;
+            const threshold =
+                onEndReachedThreshold != null ? onEndReachedThreshold * contentLength : 2;
```
sertony: Can you add a comment that here we replicate the VirtualizedList behaviour, and the default value of `2` is taken from the SectionList documentation:
https://reactnative.dev/docs/sectionlist#onendreachedthreshold

Btw, in my understanding a threshold should be calculated a bit differently according to the documentation, i.e the `onEndReachedThreshold` is just a multiplier to the visible length of the list (which is equal to windowHeight):
```suggestion
            const threshold = windowHeight * (onEndReachedThreshold ?? 2);
```

kit/scrolls/src/Lists/UIMasonryList.tsx
```diff
@@ -335,6 +339,28 @@ function useVirtualization<Item>(
         [initialWindowHeight, onLayoutProp, setWidth],
     );
 
+    const maybeCallOnEndReached = React.useCallback(
+        (contentLength: number, visibleLength: number) => {
```
sertony: Probably it's better to rename the `visibleLength` variable, since one could think that it's a "length" of the component itself. How about `scrolledContentLength` or alike?

kit/scrolls/src/Lists/UIMasonryList.tsx
```diff
@@ -403,11 +429,21 @@ function useVirtualization<Item>(
             lastTopCellIndex.current = top;
             lastBottomCellIndex.current = bottom;
 
+            maybeCallOnEndReached(event.nativeEvent.contentSize.height, y + windowHeight);
```
sertony: To make the code look cleaner and more consistent I suggest you to get the contentHeight value in [367 line](https://github.com/tonlabs/UIKit/blob/76de4ea0421dad332d9ed7b467ba7c6844c5a956/kit/scrolls/src/Lists/UIMasonryList.tsx#L367..L370), since

kit/scrolls/src/Lists/UIMasonryList.tsx
```diff
@@ -243,10 +243,14 @@ function useVirtualization<Item>(
     onLayoutProp: ScrollViewProps['onLayout'],
     onScrollProp: ScrollViewProps['onScroll'],
     setWidth: (w: number) => void,
+    onEndReached?: (info?: { distanceFromEnd: number }) => void,
+    onEndReachedThreshold?: number,
 ) {
     const lastTopCellIndex = React.useRef(0);
     const lastBottomCellIndex = React.useRef(0);
 
+    const sentEndForContentLength = React.useRef(0);
```
sertony: Can we rename the variable? I honestly don't understand what is `sent` in terms of variables...
Is it kinda `contentLenghtOnEndReached`?

kit/scrolls/src/Lists/UIMasonryList.tsx
```diff
@@ -335,6 +339,28 @@ function useVirtualization<Item>(
...
+                contentLength !== sentEndForContentLength.current
+            ) {
+                // Only call onEndReached once for a given content length
+                sentEndForContentLength.current = contentLength;
+                onEndReached({ distanceFromEnd });
+            } else if (distanceFromEnd > threshold) {
+                // If the user scrolls away from the end and back again cause
+                // an onEndReached to be triggered again
+                sentEndForContentLength.current = 0;
+            }
```
sertony: Let's avoid unwanted reassignment
```suggestion
            } else if (sentEndForContentLength.current != 0 && distanceFromEnd > threshold) {
                // If the user scrolls away from the end and then returns back again,
                // it should cause the `onEndReached` callback to be triggered again
                sentEndForContentLength.current = 0;
            }
```

ghost:
@sertony I changed the values a little, so now onEndReached works exactly as it should.

sertony (approved):

vetkamak (approved):

PR #1340, opened 2021-10-28
Add lightbox gif for android.

SAnatoliiS:


kit/media/android/src/main/java/tonlabs/uikit/media/HDuplicateImageView.java
```diff
@@ -2,38 +2,26 @@
...
 
     private final ThemedReactContext context;
     private final UIManagerModule uiManagerModule;
-    private final FrescoModule frescoModule;
 
     public HDuplicateImageView(Context context) {
         super(context);
         this.context = (ThemedReactContext) context;
         uiManagerModule = this.context.getNativeModule(UIManagerModule.class);
-        frescoModule = this.context.getNativeModule(FrescoModule.class);
```
sertony, line 36: So it looks like `fresco` wasn't really needed? (Do you remember why @savelichalex has decided to deal with it?)

sertony (approved):

kit/media/android/src/main/java/tonlabs/uikit/media/HDuplicateImageView.java
SAnatoliiS: Yes, it wasn't necessary. I don't think there was a purposeful use of this package. Perhaps that solution seemed more obvious.

ghost (approved):

PR #1341, opened 2021-10-28
Calendar improvement: first day of week by locale

ghost:


casts/dateTimePicker/src/UIDateTimePicker/Calendar/useCalendar.ts
```diff
@@ -100,13 +97,20 @@ export function useDaysCalendar() {
...
-        () => Object.values(uiLocalized.DateTimePicker.dayNamesShort),
-        [],
-    );
+    // eslint-disable-next-line radix
+    const firstDayOfWeek: number = React.useMemo(() => uiLocalized.getFirstDayOfWeek(), []);
+
+    const dayNamesShort = React.useMemo(() => {
+        const arr = Object.values(uiLocalized.DateTimePicker.dayNamesShort);
+        if (arr && firstDayOfWeek) {
+            const arr2 = arr.shift();
```
SAnatoliiS: `arr2` is not an array. Can you find a more appropriate name?

casts/dateTimePicker/src/UIDateTimePicker/Calendar/useCalendar.ts
```diff
@@ -100,13 +97,20 @@ export function useDaysCalendar() {
...
+            arr2 && arr.push(arr2);
+        }
```
sertony: What if in some mystic country the firstDatOfTheWeek is Wednesday? =)
I believe that you need to apply the "shift" as much time as it's required

localization/android/src/main/java/tonlabs/uikit/localization/UIKitLocalizationModule.java
```diff
@@ -52,6 +53,10 @@ public String getName() {
         dates.put("localePattern", dateInfo.getLocalePattern());
         dates.put("components", components);
 
+        Calendar calendar = Calendar.getInstance();
+
+        dates.put( "dayOfWeek", calendar.getFirstDayOfWeek());
```
sertony: Prettier should have fixed it
```suggestion
        dates.put("dayOfWeek", calendar.getFirstDayOfWeek());
```

localization/src/utils/getters.ts
```diff
@@ -52,9 +52,15 @@ export function getDateFormatInfo(): DateFormatInfo {
     localePattern = `${localePattern}${symbols[components[1]]}`;
     localePattern = `${localePattern}${separator}${symbols[components[2]]}`;
 
+    // Determining the first day of the week by locale when using library can consume too much space,
+    // but without it it will be inaccurate,
+    // so for simplicity we will define that on the web all weeks start on Monday, as the most common day.
```
sertony: I'm not sure if this is very correct behaviour.

I have one suggestion for that which @savelichalex might not like, but it can by applied only for web ;)

1) We can move this file out of `core` to the `localization` package:
https://github.com/tonlabs/UIKit/blob/development/packages/core/src/configs/DayJSConfig.js
2) And extend the `dayjs` by weekday plugin only for web (in order to use it here as `dayjs().weekday()`:
https://day.js.org/docs/en/plugin/weekday

P.S. **Later** we might want to clean up this part in order to avoid using all these plugins on mobile as we unfortunately do right now. And leave them only for web. @savelichalex FYI

SAnatoliiS (approved):

localization/tsconfig.json
```diff
@@ -5,5 +5,5 @@
         "experimentalDecorators": true,
         "resolveJsonModule": true
     },
-    "include": ["**/*.ts", "**/*.tsx", "**/*.json"]
+    "include": ["**/*.ts", "**/*.tsx", "**/*.json", "src/DayJSConfig.js"]
```
sertony: Oh, let's switch `DayJSConfig.js` to `DayJSConfig.ts` instead....

localization/src/index.ts
```diff
@@ -1,3 +1,5 @@
+import './DayJSConfig.js';
```
sertony:
```suggestion
import './DayJSConfig';
```

localization/src/DayJSConfig.js
```diff
@@ -0,0 +1,35 @@
+import { Platform } from 'react-native';
+// @flow
```
sertony: Consider removing @flow (it won't be needed for TS)

localization/src/index.ts
sertony: Also, since you've moved it from @packages/core, let's kill it there (as it's not needed anymore)

localization/src/utils/getters.ts
```diff
@@ -52,9 +54,17 @@ export function getDateFormatInfo(): DateFormatInfo {
...
+    // @ts-expect-error
+    const dayOfWeek = dayjs().weekday(0);
```
sertony: Why it has an error? What about the default value (for non-web platforms)?

localization/src/utils/getters.ts
```diff
@@ -52,9 +54,17 @@ export function getDateFormatInfo(): DateFormatInfo {
...
+    console.log(dayOfWeek);
```
sertony: Consider removing the log (or provide a better description for it)

sertony (changes requested):
I've just realized that `changeLocaleInfo` method will become breaking for Surf (as in Surf it doesn't have dayOfWeek property)

localization/src/service.ts
```diff
@@ -206,6 +206,10 @@ export class LocalizationService<T> extends (LocalizedStringsService as Localize
         return Language.En;
     }
 
+    getFirstDayOfWeek = (): number => {
+        return this.localeInfo.dates.dayOfWeek || 1;
```
savelichalex: Not sure about 1 as it's usually 0 AFAIR

localization/src/utils/getters.ts
savelichalex: All that fight for bundle size was useless...

localization/src/service.ts
ghost: @savelichalex  0 is sunday, usually is monday - 1

localization/src/utils/getters.ts
ghost: @sertony okay, this PR still in work :)

sertony: Ok, I've marked it as draft then ;)

localization/src/service.ts
sertony: I guess in USA (and some other countries) it's more common to start the week from Sunday. Agree with @savelichalex

ghost: @sertony @savelichalex Are we focusing on America and eastern countries?
https://en.wikipedia.org/wiki/Week#:~:text=While%2C%20for%20example%2C%20the%20United,first%20day%20of%20the%20week.

savelichalex: It's just what default in a browser, I would better stick to that

ghost: @savelichalex which browser? By ISO first is monday

savelichalex: chrome, btw you can use `Data` method as a fallback

savelichalex (approved):

sertony (approved):

PR #1344, opened 2021-11-01
Fix signing box enter

ghost:


SAnatoliiS (approved):

kit/inputs/src/UIKeyTextView.tsx
```diff
@@ -93,12 +82,22 @@ export function useKeyTextView(
         };
     }, [inputHasValue, hasInvalidChars, hasProperLength, isFocused]);
 
+    const onSubmitEditing = React.useCallback(
+        e => {
+            if (success) {
+                onKeyPress(e);
```
sertony: Does `e` handles properly here in `onKeyPress`, i.e. does it work well on web?

sertony (approved):

kit/inputs/src/UIKeyTextView.tsx
```diff
@@ -122,10 +121,12 @@ export const UIKeyTextView = React.forwardRef<UIMaterialTextViewRef, UIKeyTextVi
                 onFocus={onFocus}
                 onBlur={onBlur}
                 onChangeText={onChangeText}
-                onKeyPress={onKeyPress}
+                onSubmitEditing={onSubmitEditing}
                 helperText={helperText}
                 success={success}
                 error={error}
+                multiline={false}
```
sertony: To be honest I'd like this component to be **multiline** (in order to be able to catch the typo if any, otherwise it's very difficult to do)

Screenshot 2021-11-02 at 13 05 35

But to make it possible, IMHO it's better to show "Send" blue button to the right, instead of the X one. (as we do for UIAddressInput).

In this case no `onSubmitEditing` would be needed. And `onKeyPress` will only be used for Web purposes (to let the user press "Enter")

sertony (changes requested):

kit/inputs/src/UIKeyTextView.tsx
```diff
@@ -123,9 +125,11 @@ export const UIKeyTextView = React.forwardRef<UIMaterialTextViewRef, UIKeyTextVi
                 onBlur={onBlur}
                 onChangeText={onChangeText}
                 onKeyPress={onKeyPress}
+                onDone={onDone}
```
sertony: I don't think it's a good idea to add an extra `onDone` prop inside of a pure `UIMaterialTextView` component. 
I'd suggest to add this ActionButton component inside of `UIKeyTextView` only

stories/browser/src/UIAddressInput.tsx
```diff
@@ -2,7 +2,12 @@ import * as React from 'react';
 import { Platform, StyleSheet, TextInput } from 'react-native';
 
 import { ChatInputContainer, useChatInputValue } from '@tonlabs/uistory.chats';
-import { UITextView, useAutogrowTextView, UIInputAccessoryView } from '@tonlabs/uikit.inputs';
+import {
+    ActionButton,
```
sertony: I also think that `ActionButton` should be moved to another package, but not in `uikit.input`. @savelichalex FYI

sertony (changes requested):

kit/inputs/src/UIKeyTextView.tsx
ghost: @sertony But UIMaterialTextView have useMaterialTextViewChildren which is made especially for similar buttons, and thats why I pass onDone here

stories/browser/src/UIAddressInput.tsx
ghost: @sertony ActionButton essentially the same as useClearButton, a little more independent but still only for input components, thats why I put it here. What place will be better?
May be it's better to made another independent component that can be clear button or action button. And then it may be placed in uikit.controls for example. For now everything, for no particular reason, lies in different places.

savelichalex: I'm OK with leaving it in `uikit.input`

kit/inputs/src/UIKeyTextView.tsx
savelichalex: I would say that sth is actually wrong with `useMaterialTextViewChildren`, it shouldn't actually have `onDone` argument, it rather should be passed as child component, as it was supposed to be. Solve bigger amount of cases > solve one case.

sertony: @wearedusa I've checked the code of UIMaterialTextView. You can pass the `.Icon` element as a children of the input, as it's done here: https://github.com/tonlabs/UIKit/blob/676a56bf3759edebab71b2e76486f6ad9b108cc6/casts/addressText/src/UIAddressTextView.tsx#L155..L162

Hence, you won't need any unwanted `onDone` props and even ActionButton ;)

kit/inputs/src/UIKeyTextView.tsx
```diff
@@ -126,7 +127,22 @@ export const UIKeyTextView = React.forwardRef<UIMaterialTextViewRef, UIKeyTextVi
...
                 error={error}
-            />
+                multiline
+            >
+                <UIMaterialTextView.Icon
+                    testID="send_btn"
+                    source={UIAssets.icons.ui.buttonMsgSend}
+                    onPress={onKeyPress}
+                    style={styles.icon}
+                />
```
sertony: Probably we shouldn't render it until we don't have a `success` stage...

savelichalex (approved):

sertony (approved):

kit/inputs/src/UIMaterialTextView/useMaterialTextViewChildren.tsx
```diff
@@ -84,10 +84,11 @@ export function useMaterialTextViewChildren(
     isFocused: boolean,
     isHovered: boolean,
     clear: () => void,
+    success?: boolean,
 ) {
     const clearButton = useClearButton(inputHasValue, isFocused, isHovered, clear);
 
-    if (clearButton) {
+    if (clearButton && !success) {
```
savelichalex: Please leave a comment with a case where this is needed, because it doesn't look like a general case and probably we should change it in a future to sth more generic, I just don't know right now how it should be

ghost: @savelichalex done

sertony: How about renaming the parameter? Into something like `hideClearButton`? As it doesn't bring any context like success (which might be different)

kit/inputs/src/UIMaterialTextView/useMaterialTextViewChildren.tsx
```diff
@@ -84,10 +84,15 @@ export function useMaterialTextViewChildren(
...
     clear: () => void,
+    success?: boolean,
 ) {
     const clearButton = useClearButton(inputHasValue, isFocused, isHovered, clear);
-
-    if (clearButton) {
+    
+    /**
+     * If clearButton is visible it blocks displaying child icons.
+     * So we show it if the input is invalid.
```
savelichalex:
```suggestion
     * If clearButton is visible it blocks displaying child icons.
     * So we show it if the input is invalid.
     *
     * It was made for UIKeyTextView, as for the signing box interface in the browser
     * we want to show the icon of submitting, when a key is valid, instead of a clear button
```

kit/inputs/src/UIMaterialTextView/UIMaterialTextView.tsx
```diff
@@ -381,10 +390,14 @@ const UIMaterialTextViewSimple = React.forwardRef<
...
-    const clear = React.useCallback(() => {
-        clearInput();
+    const clearInput = React.useCallback(() => {
+        clearProp();
         resetInputHeight();
-    }, [clearInput, resetInputHeight]);
+    }, [clearProp, resetInputHeight]);
+
+    const clear = React.useMemo(() => {
+        return success ? undefined : () => clearInput();
```
savelichalex: So what's the point of the last commit if it does the same? btw `() => clearInput()` can be safely changed to `clearInput`

kit/inputs/src/UIMaterialTextView/useClearButton.tsx
```diff
@@ -11,10 +11,10 @@ export function useClearButton(
     inputHasValue: boolean,
     isFocused: boolean,
     isHovered: boolean,
-    clear: () => void,
+    clear?: () => void,
```
savelichalex, line 14: Without current context it can be so not obvious how clear button can be created without it IMO

kit/inputs/src/UIMaterialTextView/UIMaterialTextView.tsx
ghost: @savelichalex This is @sertony idea

savelichalex: @sertony I didn't get the idea

kit/inputs/src/UIMaterialTextView/useClearButton.tsx
ghost: @sertony FYI

sertony: That's the thing, that the clear button SHOULD NOT be created without it ;) I like the idea

sertony (approved):

kit/inputs/src/UIMaterialTextView/UIMaterialTextView.tsx
```diff
@@ -285,11 +285,11 @@ const UIMaterialTextViewFloating = React.forwardRef<
     UIMaterialTextViewRef,
     UIMaterialTextViewCommonProps
 >(function UIMaterialTextViewFloatingForwarded(props: UIMaterialTextViewCommonProps, passedRef) {
-    const { label, onLayout, children, onHeightChange, ...rest } = props;
+    const { label, onLayout, children, onHeightChange, success, ...rest } = props;
     const ref = React.useRef<TextInput>(null);
     const {
         inputHasValue,
-        clear: clearInput,
+        clear: clearProp,
```
savelichalex: Btw it's not a prop, it's not passed as a prop anywhere

ghost: ah

sertony:
Converted to Draft. This task needs to be solved together with the designer now!

sertony:
Well, @eteslov has provided an updated design for the button:
https://www.figma.com/file/bvGrAoGcpwwh1t8YpTfnq5/Surf-Workflow?node-id=7784%3A38

PR #1345, opened 2021-11-01
BubblePlainText url parsing improvement

ghost:


stories/chats/src/constants.ts
```diff
@@ -2,6 +2,12 @@ import { Platform } from 'react-native';
 
 const GRID_LINE_HEIGHT = 4;
 
+export const RegExpConstants = {
+    domain: /.+\/\/|www.|\/.+/g,
```
sertony: This is not needed?

stories/chats/src/constants.ts
```diff
@@ -2,6 +2,12 @@ import { Platform } from 'react-native';
...
+    protocol: /^https?:\/\//,
+    url: /(https?:\/\/|www\.)[-а-яА-Яa-zA-Z0-9@:%._\+~#=]{1,256}\.[a-z]{2,6}\b([-а-яА-Яa-zA-Z0-9@:%_\+.~#?&\/=]*[-а-яА-Яa-zA-Z0-9@:%_\+~#?&\/=])*/i,
```
sertony, line 7: Can you also add a comment why we've changed the URL regex here (since before we used the default one from `react-native-parsed-text`: https://github.com/taskrabbit/react-native-parsed-text/blob/fded31e2012959d1d4c3ade0e47e4d826e732fd0/src/ParsedText.js#L26)

ghost: @sertony done!
https://github.com/tonlabs/UIKit/pull/1345/files#diff-cd7ebb53e00bb93de58a5ea8fd899dfad0402a99b8ebc2b4a8fe2ec55a3cea7cR210-R215

sertony (approved):

SAnatoliiS (approved):

PR #1355, opened 2021-11-09
UICountryPicker: Fixed android keyboard

ghost:
https://www.notion.so/tonlabs/Surf-Android-Country-picker-Problems-with-the-opened-keyboard-c218b14465cb401f908cbbac1a6ac0d7


Example/android/app/build.gradle
```diff
@@ -240,6 +240,8 @@ dependencies {
         exclude group:'com.facebook.flipper'
     }
 
+    compile project(":react-native-android-keyboard-adjust")
```
savelichalex: hm, why? it doesn't support auto-linking?

casts/countryPicker/src/UICountryPicker/CountryPicker.tsx
```diff
@@ -133,7 +148,12 @@ export function CountryPicker({
                     </UILabel>
                     <View style={styles.sideHeaderView} />
                 </View>
-                <UISearchBar returnKeyType="done" value={search} onChangeText={setSearch} />
+                <UISearchBar
+                    returnKeyType="done"
+                    value={search}
+                    onChangeText={setSearch}
+                    onFocusChanged={setFocused}
```
savelichalex: Not sure that you need to create a new prop though, also not sure that you should use state there, maybe just pass `onFocus` and `onBlur` and call methods there?

casts/countryPicker/src/UICountryPicker/CountryPicker.tsx
```diff
@@ -121,6 +129,13 @@ export function CountryPicker({
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
+    const hideKeyboard = React.useCallback(() => {
+        // react-native-android-keyboard-adjust bug:
```
savelichalex, line 144: Not sure that it's a bug of the lib, I suspect it's sth with `keyboardDismissMode`

casts/countryPicker/src/UICountryPicker/CountryPicker.tsx
```diff
@@ -121,6 +129,13 @@ export function CountryPicker({
...
+        // Keyboard doesn't want to hide on Android
+        // so we have to forcibly hide the keyboard
+        isAndroid && focused && Keyboard.dismiss();
```
savelichalex: Btw AFAIK you could safely call `Keyboard.dismiss()` without checking whether it's focused or not

Example/android/app/build.gradle
ghost: @savelichalex It seems so.

casts/countryPicker/src/UICountryPicker/CountryPicker.tsx
ghost: @savelichalex onBlur also does not work on my android.

casts/countryPicker/src/UICountryPicker/CountryPicker.tsx
ghost: @savelichalex I wrote incorrectly. In fact, sometimes it closes, but with a long delay, when I use keyboardDismissMode="on-drag"

casts/countryPicker/src/UICountryPicker/CountryPicker.tsx
ghost: @savelichalex okay it work again

ghost: @savelichalex sometimes

casts/countryPicker/src/UICountryPicker/CountryPicker.tsx
savelichalex: Have you tried https://reactnative.dev/docs/scrollview#keyboardshouldpersisttaps?

Example/android/app/build.gradle
sertony: Same for Surf, they do not support it indeed.

casts/countryPicker/package.json
```diff
@@ -34,16 +34,17 @@
...
-        "@tonlabs/uikit.popups": "^2.3.0",
+        "@tonlabs/uikit.core": "^2.3.0",
         "@tonlabs/uikit.layout": "^2.3.0",
         "@tonlabs/uikit.media": "^2.3.0",
-        "@tonlabs/localization": "^2.3.0",
+        "@tonlabs/uikit.popups": "^2.3.0",
+        "@tonlabs/uikit.scrolls": "^2.3.0",
+        "@tonlabs/uikit.themes": "^2.3.0",
         "fuse.js": "^3.4.6",
+        "react-native-android-keyboard-adjust": "^1.2.0",
```
sertony: This one should be a `peerDependency` (as well as devDependency), not a regular dependency!

casts/countryPicker/src/UICountryPicker/CountryPicker.tsx
ghost: yes

SAnatoliiS (approved):

Example/android/app/build.gradle
savelichalex: btw I believe it should be `implementation` instead of `compile`, AFAIR it should complain about it in AS and during a build

casts/countryPicker/package.json
```diff
@@ -66,7 +66,8 @@
         "react-native-gesture-handler": "^1.10.3",
         "react-native-reanimated": "^2.2.1",
         "react-native-redash": "^16.0.11",
-        "react-native-svg": "^12.1.1"
+        "react-native-svg": "^12.1.1",
+        "react-native-android-keyboard-adjust": "^1.2.0"
```
sertony, line 71: This should also be added to the `devDependencies` (otherwise TS / ESLint should actually complain...)

When you add anything in `peerDependencies` it's always a good practice to have them in `devDependencies` too

casts/countryPicker/types/index.d.ts
```diff
@@ -1 +1,2 @@
 declare module '@tonlabs/uikit.core';
+declare module 'react-native-android-keyboard-adjust';
```
sertony, line 2: Oh, here is why TS didn't complain, I wonder why ESLint is silent in this case)

sertony (approved):

casts/countryPicker/src/UICountryPicker/CountryPicker.tsx
```diff
@@ -104,6 +107,10 @@ export function CountryPicker({
     }, [search]);
 
     React.useEffect(() => {
+        // We should use react-native-android-keyboard-adjust
+        // to fix android keyboard work with UIBottomSheet
+        isAndroid && AndroidKeyboardAdjust.setAdjustNothing();
```
sertony: Well, I've tested it. 

```
AndroidKeyboardAdjust.setAdjustNothing()
```
The script above makes the keyboard `not adjusted` throw all the screens.

And when I close this modal and open the chat, they keyboard is not adjusted (as RESIZE) anymore.

Consider returning the following function on unmount with some proper comment
```
() => isAndroid && AndroidKeyboardAdjust.setAdjustXXXX(); // where XXX is a previous state
```

sertony (changes requested):

casts/countryPicker/src/UICountryPicker/CountryPicker.tsx
sertony: Actually it's better to find the previous state and return it on unmount!

ghost: @sertony Yes, you are right. I rewrote this part, but not onmount/unmount, but when changing `visible`, this is more suitable in this case.

casts/countryPicker/src/UICountryPicker/CountryPicker.tsx
```diff
@@ -121,6 +124,26 @@ export function CountryPicker({
...
+    React.useEffect(() => {
+        if (isAndroid) {
+            // We should use react-native-android-keyboard-adjust
+            // to fix android keyboard work with UIBottomSheet
+            // also we have to check if component visible
+            // and only if it showing call setAdjustNothing
+            // because in other cases we may need the default adjust - setAdjustPan
+            visible
+                ? AndroidKeyboardAdjust.setAdjustNothing()
+                : AndroidKeyboardAdjust.setAdjustPan();
```
sertony: Can we somehow find the initial "mode"? Since I'm not sure that we need to return into `pan` mode, I'd rather return into `Resize` (as the most used in Surf)

ghost: @sertony As I see resize does not work in chat(setAdjustPan work), I will look for something

sertony (approved):

savelichalex (approved):

PR #1356, opened 2021-11-10
Add react-native-video.

SAnatoliiS:


savelichalex (approved):

kit/media/src/UIVideo/UIVideo.native.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import RNVideo from 'react-native-video';
```
sertony, line 2: Since we do not need to use `react-native-video` for web, you can exclude it from the final bundling by setting an alias for it in **webpack.js**, i.e.:
```
'react-native-video': 'react', // Hack in order not to load
```

kit/media/src/UIVideo/README.md
```diff
@@ -0,0 +1,24 @@
+# UIVideo
```
sertony, line 1: 👍 Glad you've added it!

SAnatoliiS:
Should we put this component in a separate `cast` package? @sertony @savelichalex

kit/media/src/UIVideo/hooks.ts
```diff
@@ -0,0 +1,45 @@
...
+                height: width / aspectRatio,
+            };
+        }
+        if (width && height) {
+            if (height > width / aspectRatio) {
+                return {
+                    width,
+                    height: width / aspectRatio,
+                };
+            }
```
sertony: Very small optimization suggestion :)
```suggestion
            const aspectedHeight = width / aspectRatio;
            if (aspectedHeight < height) {
                return {
                    width,
                    height: aspectedHeight,
                };
            }
```

P.S. I was thought that `<` is always better than `>` in terms of Assembly optimization, just an interesting fact ;)

sertony (approved):

sertony:
> Should we put this component in a separate `cast` package? @sertony @savelichalex

As for my opinion, I'd like having it as a core video component inside of `uikit.media`.

sertony (approved):

savelichalex (approved):

Example/package.json
```diff
@@ -95,7 +95,7 @@
         "react-native-simple-popover": "git+https://github.com/tonlabs/react-native-simple-popover.git",
         "react-native-status-bar-height": "1.0.1",
         "react-native-svg": "^12.1.0",
-        "react-native-video": "^5.2.0",
+        "react-native-video": "git+https://github.com/r0b0t3d/react-native-video.git#feature/android-fullscreen",
```
sertony: Why so? This is an old repo which hasn't been updated for years.... :(

sertony (changes requested):
To be honest I'd better miss some feature than integrate so outdated and not maintained package in our codebase.

Example/package.json
SAnatoliiS: It's all for fullscreen mode on android. This is a critical feature

SAnatoliiS: I changed it to `tonlabs/react-native-video`

sertony (approved):

PR #1357, opened 2021-11-11
Get rid of legacy

ghost:
https://github.com/tonlabs/TON-Surf/pull/2894


SAnatoliiS (approved):

sertony (approved):

packages/components/src/UICompatibilityView/index.web.js
```diff
@@ -1,17 +0,0 @@
-// @flow
-import React from 'react';
-import { Platform, StyleSheet, Text, View } from 'react-native';
-// TODO: we have `mobile-detect` in UIDevice
-// maybe use only one lib for
-import Bowser from 'bowser';
```
sertony, line 6: Btw, we can remove the `bowser` from dependencies then

packages/components/src/UICompatibilityView/index.web.js
```diff
@@ -1,17 +0,0 @@
...
-
-import UICompatibilityViewCommon from './UICompatibilityViewCommon';
-
-export default class UICompatibilityView extends React.Component<*> {
-    render() {
-        if (!Bowser.msie) {
```
sertony, line 12: @savelichalex FYI, this check was used to display a message for old versions of IE that we do not support their browser

packages/components/src/UICompatibilityView/UICompatibilityViewCommon.js
```diff
@@ -1,34 +0,0 @@
...
-        justifyContent: 'center',
-        alignItems: 'center',
-    },
-});
-
-export default class UICompatibilityViewCommon extends UIComponent<null, null> {
-    renderNotCompatible() {
-        return (
-            <Text style={UITextStyle.primaryBodyRegular}>
-                {uiLocalized.WeAreSorryButYourBrowserVersionIsNotCompatible}
```
sertony, line 22: If we don't need it anymore, you can clean up this translation as well

savelichalex (approved):

packages/components/src/UICompatibilityView/UICompatibilityViewCommon.js
ghost: @sertony done!

packages/components/src/UICompatibilityView/index.web.js
ghost: @sertony done!

PR #1358, opened 2021-11-12
UISkeleton & useImageSize hook updates

vetkamak:


SAnatoliiS (approved):

stories/chats/src/BubbleMedia/hooks.tsx
```diff
@@ -162,9 +164,19 @@ export const useImageSize = (data: string | null) => {
...
+        if (originalSize && originalSizeRef.current) {
+            const newImageSize = getImageSize(
+                originalSizeRef.current?.width,
+                originalSizeRef.current?.height,
+                maxImageSize.width,
+                maxImageSize.height,
+            );
+            setImageSize(newImageSize);
+            return;
+        }
```
sertony: Looks a little like a double-code from above. Possibly we are to move `originalSize` as a dependency there.... BUT I don't see where the `originalSize` value is really used here..

vetkamak: `originalSize` value is used as the initial for originalSizeRef, so when we already know the image's dimensions we won't need to use Image.getSize() to get them

sertony (approved):

stories/chats/src/BubbleMedia/hooks.tsx
```diff
@@ -140,9 +140,11 @@ export const useImages = (
     }, [data, preview, imageRef, imageSize, onErrorCallback, onLoadCallback]);
 };
 
-export const useImageSize = (data: string | null) => {
+export const useImageSize = (data: string | null, originalSize?: ImageSize) => {
     const [imageSize, setImageSize] = React.useState<ImageSize | null>(null);
-    const originalSizeRef = React.useRef<ImageSize | null>(null);
+    const originalSizeRef = React.useRef<ImageSize | null>(
+        originalSize === undefined ? null : originalSize,
```
savelichalex:
```suggestion
        originalSize ?? null
```

kit/layout/src/Skeleton/Skeleton.tsx
```diff
@@ -190,13 +195,18 @@ export function UISkeleton({
 
     const { isVisible, crossDissolveProgress } = useCrossDissolve(visible);
 
+    const containerStyles = [styles.container, styleProp];
+
     return (
-        <UIBackgroundView style={[styles.container, styleProp]} onLayout={onLayout}>
+        <View
+            style={persistStyles ? containerStyles : isVisible ? containerStyles : null}
```
savelichalex: Let's extract it to `useMemo`, nested ternaries are hard to reason about :)

sertony (approved):

sertony (approved):

kit/layout/src/Skeleton/Skeleton.tsx
```diff
@@ -190,13 +195,21 @@ export function UISkeleton({
 
     const { isVisible, crossDissolveProgress } = useCrossDissolve(visible);
 
+    const containerStyles = React.useMemo(() => {
+        const containerStyles = [styles.container, styleProp];
+        if (persistStyles) {
+            return containerStyles;
+        }
+        return isVisible ? containerStyles : null;
+    }, [isVisible, persistStyles]);
```
sertony: @savelichalex have just pointed out that this is a breaking change in API!

I'd better agree with Aleksey that the SkeletonView should remain to be a View without any style related logic.

sertony (changes requested):

sertony (approved):

savelichalex (approved):

PR #1361, opened 2021-11-17
SpinnerOverlay refactoring

ghost:


savelichalex (approved):

sertony:
As I see the legacy UISpinnerOverlay is still used by `navigation_legacy` and `legacy` packages (e.g. UISharedComponents, which seem to be safely removed from UIKit, and UIController where it seems to be used a little still...)

ghost:
@sertony yes

casts/spinnerOverlay/src/SpinnerOverlay/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+    titleContent: string;
+    textContent: string;
+};
+
+const waveConfig = {
+    size: 40,
+    count: 2,
+};
+
+export function UISpinnerOverlay({ visible = false, titleContent, textContent }: Props) {
```
sertony: I've checked the codebase and noticed that the `titleContent` and `textContent` are not really used. Actually it's used only in UIController in `setSpinnerTitleContent` and `setSpinnerTextContent` methods, which also not used at all anywhere in Surf ;)

casts/spinnerOverlay/src/SpinnerOverlay/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+const waveConfig = {
+    size: 40,
+    count: 2,
+};
+
+export function UISpinnerOverlay({ visible = false }: Props) {
+    const renderSpinner = React.useMemo(() => {
+        return (
+            <View testID="spinner_loader" style={[styles.container]} key={`spinner~${Date.now()}`}>
+                <View style={styles.background}>
```
sertony: It looks like we can now use only one View instead of two wrapped one in each other...
What do you think?

casts/spinnerOverlay/src/SpinnerOverlay/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+    },
+    background: {
+        position: 'absolute',
+        top: 0,
+        bottom: 0,
+        left: 0,
+        right: 0,
+        justifyContent: 'center',
+        alignItems: 'center',
+    },
```
sertony: Probably we can have a single container style now only (as suggested above)
```suggestion
    container: {
        position: 'absolute',
        top: 0,
        bottom: 0,
        left: 0,
        right: 0,
        backgroundColor: 'rgba(0,0,0,0.15)',
        justifyContent: 'center',
        alignItems: 'center',
    },
```

sertony (approved):

PR #1364, opened 2021-11-18
fix spinner issues

sertony:


Example/package.json
```diff
@@ -79,7 +79,6 @@
         "react-native-fast-image": "8.3.4",
         "react-native-gesture-handler": "1.10.3",
         "react-native-image-picker": "git+https://github.com/FactorBench/react-native-image-picker.git#develop",
-        "react-native-indicators": "git+https://github.com/tonlabs/react-native-indicators.git",
```
savelichalex, line 82: Do you really want to delete it from Example, as I see it's a peerDep

sertony: It's not a peer (as it's a pure JS module)

savelichalex (approved):

PR #1369, opened 2021-11-23
Add CollectionCard.

SAnatoliiS:


casts/cards/src/UICollectionCard/UICollectionCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+import type { UICollectionCardProps } from './types';
+import { QuickView } from './QuickView';
+import { Title } from './Title';
+import { Badge } from './Badge';
+import { UIConstant } from '../constants';
+
+const linearGradientStart = 'rgba(16, 19, 21, 0.8)';
+const linearGradientEnd = 'rgba(16, 19, 21, 0.2)';
+
+export const UICollectionCard: React.FC<UICollectionCardProps> = ({
```
savelichalex: I like the naming! :)

Though please make it with a named `function` as it works better for development (please consider that we must use it for all components instead of anonymous functions)

casts/cards/src/UICollectionCard/UICollectionCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+        borderRadius: UIConstant.uiCollectionCard.borderRadius,
+        backgroundColor: theme[ColorVariants.BackgroundOverlay],
+        padding: UILayoutConstant.contentOffset,
+        justifyContent: 'space-between',
+        aspectRatio: 1,
+        overflow: 'hidden',
+    },
+    imageList: {
+        ...StyleSheet.absoluteFillObject,
+        zIndex: -20,
```
savelichalex: Why do you need `zIndex`? AFAIR it works badly on mobile, so we tend not to use it, better to rely on hierarchy position (lower position in hierarchy place it above)

casts/cards/src/UICollectionCard/types.ts
```diff
@@ -0,0 +1,25 @@
...
+    imageSourceList?: ImageSourcePropType[];
+    style: ViewStyle;
+};
+
+export type TitleProps = {
+    title?: string;
+};
+
+export type BadgeProps = {
+    badge?: number | React.ReactElement<UIImageProps>;
```
savelichalex: what do you think instead of `React.ReactElement<UIImageProps>` pass `UIImageProps['source']`, and set all needed styles (especially width and height) themselves? This way all a developer need to do is to just import from assets and pass it

casts/cards/src/UICollectionCard/UICollectionCard.tsx
SAnatoliiS: I did so. But something on the web is not working correctly. I had to explicitly specify the zIndex, but the hierarchy is also complied. It works fine on mobile and web both.

SAnatoliiS:
Снимок экрана 2021-11-24 в 12 43 02

Снимок экрана 2021-11-24 в 12 43 12

Снимок экрана 2021-11-24 в 12 44 02

savelichalex (approved):

casts/cards/src/UICollectionCard/types.ts
```diff
@@ -0,0 +1,27 @@
+import type { ViewStyle, ImageSourcePropType } from 'react-native';
+
+export type ContentType = 'Image' | 'Video' | 'Unknown';
+
+export type UICollectionCardProps = {
+    contentType: ContentType;
```
vetkamak, line 6: For now it's ok as we use a single element of the source array in the Preview, but in subsequent iterations we should be also able to process an array of heterogeneous media files formats (unfortunately we can have an image, gif and video in one collection)

vetkamak (approved):

casts/cards/src/UICollectionCard/UICollectionCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+}
+
+const useStyles = makeStyles((theme: Theme) => ({
+    skeleton: {
+        /**
+         * The radius of the skeleton has to be made a little smaller
+         * so that the corners of the content are not visible
+         */
+        borderRadius: UIConstant.uiCollectionCard.borderRadius - 2,
+    },
```
sertony: You can remove this thanks to my latest fix ;)

casts/cards/src/UICollectionCard/UICollectionCard.tsx
sertony: How did it look without applying `zIndex`? @savelichalex is right. `zIndex` is usually evil...

casts/cards/src/UICollectionCard/UICollectionCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+        aspectRatio: 1,
+        overflow: 'hidden',
+    },
+    preview: {
+        ...StyleSheet.absoluteFillObject,
+        zIndex: -20,
+    },
+    gradient: {
+        ...StyleSheet.absoluteFillObject,
+        zIndex: -10,
```
sertony: Same question as above

casts/cards/src/UIMediaCard/UIMediaCard.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+const useStyles = makeStyles((theme: Theme) => ({
+    skeleton: {
+        /**
+         * The radius of the skeleton has to be made a little smaller
+         * so that the corners of the content are not visible
+         */
+        borderRadius: UIConstant.uiMediaCard.borderRadius - 2,
+    },
```
sertony: Same consideration as above. Consider removing

casts/cards/src/UIMediaCard/UIMediaCard.tsx
```diff
@@ -0,0 +1,62 @@
...
+        borderRadius: UIConstant.uiMediaCard.borderRadius,
+        backgroundColor: theme[ColorVariants.BackgroundOverlay],
+        padding: UILayoutConstant.contentOffset,
+        justifyContent: 'flex-end',
+        aspectRatio: 1,
+        overflow: 'hidden',
+    },
+    quickView: {
+        ...StyleSheet.absoluteFillObject,
+        zIndex: -20,
```
sertony: Oh, I believe that the problem happens due to the fact that you try to wrap `2` components inside one `TouchableOpacity`, which is an anti-pattern.  This was mentioned somewhere in RN docs, that you can wrap only a single component with the Touchables...

kit/controls/src/constants.ts
```diff
@@ -13,46 +13,12 @@ export const UIConstant = {
     pillButtonHeight: 32,
     linkButtonHeight: 48,
 
-    // smallCellHeight: 24,
-    // defaultCellHeight: 48,
-
-    // minimumWidthToShowFoldingNotice: 400,
-    // minNoticeSize: 88,
-    // minNoticeIconSize: 64,
-    // maxNoticeIconSize: 88,
```
sertony, line 22: 👍

sertony (changes requested):
Read the comments above

casts/cards/src/UICollectionCard/UICollectionCard.tsx
SAnatoliiS: No, the problem is different
image

SAnatoliiS: The problem is only on the web

casts/cards/src/UICollectionCard/UICollectionCard.tsx
SAnatoliiS: image

casts/cards/src/UICollectionCard/UICollectionCard.tsx
sertony: Got it!

casts/cards/src/UIMediaCard/UIMediaCard.tsx
sertony: Already discussed

casts/cards/src/UICollectionCard/UICollectionCard.tsx
SAnatoliiS: I applied `zIndex` only for web

casts/cards/src/UICollectionCard/UICollectionCard.tsx
SAnatoliiS: I applied `zIndex` only for web

casts/cards/src/UIMediaCard/UIMediaCard.tsx
SAnatoliiS: I did a single component and applied `zIndex` only for web

sertony (approved):

PR #1371, opened 2021-11-24
Add UILink.

SAnatoliiS:


SAnatoliiS:
image
image

casts/row/src/UILink/types.ts
```diff
@@ -0,0 +1,19 @@
+import type { ImageSourcePropType } from 'react-native';
+
+export type UILinkProps = {
+    title?: string;
+    description?: string;
+    logoSource?: ImageSourcePropType;
+    iconAfterTitleSource?: ImageSourcePropType;
```
savelichalex: I don't think it should be configurable. Let's drop this prop

casts/row/src/UILink/types.ts
```diff
@@ -0,0 +1,19 @@
+import type { ImageSourcePropType } from 'react-native';
+
+export type UILinkProps = {
+    title?: string;
+    description?: string;
+    logoSource?: ImageSourcePropType;
```
savelichalex: What do you think to call it just `logo`?

casts/row/src/UILink/UILink.tsx
```diff
@@ -0,0 +1,55 @@
...
+    title,
+    description,
+    logoSource,
+    iconAfterTitleSource,
+    onPress,
+    loading,
+    testID,
+}: UILinkProps) => {
+    return (
+        <UISkeleton show={!!loading}>
```
savelichalex: I actually thought that skeleton will be around logo, title and subtitle, not around whole row. Like the point of skeletons is to show what layout will be after it's loaded

SAnatoliiS: I have separated the logo from the text block
image

savelichalex (approved):

casts/row/src/UILink/Logo.tsx
```diff
@@ -0,0 +1,40 @@
...
+import type { LogoProps } from './types';
+import { UIConstant } from '../constants';
+
+export const Logo: React.FC<LogoProps> = ({ logo, loading }: LogoProps) => {
+    if (!logo) {
+        return (
+            <View style={styles.container}>
+                <UISkeleton show={!!loading}>
+                    <View style={styles.image} />
+                </UISkeleton>
```
sertony: Am I right that the skeleton will look more like a square here than a plain circle (as the logo should look like)?

savelichalex: A logo isn't a circle anymore :)

SAnatoliiS: image

sertony (approved):

PR #1378, opened 2021-11-26
Improvement skeleton.

SAnatoliiS:


kit/layout/src/Skeleton/Skeleton.tsx
```diff
@@ -213,6 +229,25 @@ export function UISkeleton({
...
+         */
+        ...Platform.select({
+            web: {
+                top: -1,
+                bottom: -1,
+                left: -1,
+                right: -1,
+            },
+            android: {
+                top: -0.5,
```
savelichalex: Actually seems to be fragile in devices with pixel density = 1

casts/rows/src/UILink/UILink.tsx
```diff
@@ -59,6 +59,7 @@ const styles = StyleSheet.create({
     },
     textContent: {
         flex: 1,
+        borderRadius: 0,
```
savelichalex: What this for? AFAIR it's 0 by default. What is it doing?
Need to at least leave a comment, to not delete it along the way, as it seems unnecessary without the context.

SAnatoliiS: Removed, no longer required

kit/layout/src/Skeleton/Skeleton.tsx
SAnatoliiS: I agree, this solution has many drawbacks. Deleted. Used `opacity: 0` for children

kit/layout/src/Skeleton/Skeleton.tsx
```diff
@@ -193,18 +199,21 @@ export function UISkeleton({
...
-            style={[
-                styles.container,
-                { borderRadius: isVisible ? UILayoutConstant.alertBorderRadius : 0 },
-                styleProp,
-            ]}
-            onLayout={onLayout}
-        >
-            {children}
+        <View style={[styles.container, styleProp]} onLayout={onLayout}>
+            <View style={{ opacity: show ? 0 : 1 }}>{children}</View>
```
savelichalex: Does it mean that `Skeleton` isn't behave like a regular `View` anymore? For example setting `flexDirection: row` for it won't have any effect on content.

SAnatoliiS: You're right, I'm going back to the previous solution with a little revision

savelichalex (approved):

kit/layout/src/Skeleton/Skeleton.tsx
```diff
@@ -213,7 +222,13 @@ export function UISkeleton({
 const styles = StyleSheet.create({
     container: {
         position: 'relative',
+    },
+    skeletonContainer: {
         overflow: 'hidden',
+        top: -StyleSheet.hairlineWidth,
+        left: -StyleSheet.hairlineWidth,
+        bottom: -StyleSheet.hairlineWidth,
+        right: -StyleSheet.hairlineWidth,
```
sertony, line 237: That looks weird to me... Can you add a comment why do we need it?
Also, it looks like it's better to add `position: 'absolute'` here in order to ensure that the applied style contains it (without relying on `StyleSheet.absoluteFill`)

sertony (approved):

PR #1381, opened 2021-11-30
Add account row

lochmatij:


casts/rows/src/UIAccountRow.tsx
```diff
@@ -0,0 +1,121 @@
...
+    icon: {
+        width: UIConstant.uiLink.logo.size,
+        height: UIConstant.uiLink.logo.size,
+    },
+    desc: {
+        flex: 1,
+        marginRight: UILayoutConstant.normalContentOffset,
+    },
+    right: {
+        textAlign: 'right',
```
sertony: Maybe `alignItems`?

casts/rows/src/UIAccountRow.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+import { UIConstant } from './constants';
+
+export type UIAccountRowProps = {
+    testID?: string;
+    icon: UIImageProps['source'];
+    name: string;
+    description?: string;
+    balance: BigNumber;
+    currencySignIcon: any;
```
savelichalex:
```suggestion
    currencySignProps?: UICurrencySignProps;
```

What do you think about that? That way we can support all the customisations available for UICurrency sign

casts/rows/src/UIAccountRow.tsx
savelichalex: Have you tried `alignItems: 'flex-end'`? It's just I'm not sure how it can affect UICurrency :) If it works fine, then it's OK though

sertony (approved):

savelichalex (approved):

PR #1382, opened 2021-12-01
Collection card preview animation

SAnatoliiS:


kit/media/src/UIVideo/UIVideo.web.tsx
```diff
@@ -6,12 +6,38 @@ import type { UIVideoProps } from './types';
...
+            url={uri}
+            controls={controls}
+            {...dimensions}
+            loop={repeat}
+            playing={!paused}
+            muted={muted}
+            /**
+             * For apply resizeMode
+             */
+            config={{
```
savelichalex: Do you think if we should apply `React.useMemo` for it, or we can apply `React.memo` for the whole component? to decrease unneeded re-renders

SAnatoliiS:
image
image

casts/cards/src/types.ts
```diff
@@ -0,0 +1,10 @@
+import type { ImageURISource } from 'react-native';
+
+export type ContentType = 'Image' | 'Video' | 'Unknown';
+
+export type ContentSource = ImageURISource;
+
+export type Content = {
```
savelichalex: Too generic type name to export outside, need to either export it with different name with prefix, or rename it here

casts/cards/src/UIMediaCard/QuickView.tsx
```diff
@@ -1,16 +1,16 @@
...
-import { UIImage } from '@tonlabs/uikit.media';
+// import { View } from 'react-native';
+// import { UIImage } from '@tonlabs/uikit.media';
 
 import type { QuickViewProps } from './types';
+import { CollectionSlide } from '../UICollectionCard/CollectionSlide';
 
-export const QuickView: React.FC<QuickViewProps> = ({ source, style, contentType }: QuickViewProps) => {
-    if (contentType !== 'Image' ||!source) {
+export const QuickView: React.FC<QuickViewProps> = ({ content, style }: QuickViewProps) => {
```
savelichalex: Do you think we can drop this component? too much of a fragmentation if you ask me :)

casts/cards/src/UICollectionCard/usePreload/usePreload.web.tsx
```diff
@@ -0,0 +1,24 @@
...
+            return;
+        }
+
+        content.forEach((contentItem: Content): void => {
+            if (
+                contentItem.contentType === 'Image' &&
+                contentItem.source &&
+                contentItem.source.uri
+            ) {
+                RNImage.prefetch(contentItem.source.uri);
```
savelichalex: AFAIR `react-native-fast-image` have there own cache, are you sure that it's needed?

savelichalex: Oh, I see now that it's for a web. Maybe we should implement it in UIImage then?

casts/cards/src/UICollectionCard/CollectionSlide.tsx
```diff
@@ -0,0 +1,46 @@
...
+    y: 0,
+    width: 0,
+    height: 0,
+};
+
+export function CollectionSlide({ content, style }: CollectionSlideProps) {
+    const [dimensions, setDimensions] = React.useState<LayoutRectangle>(defaultDimension);
+    const onLayout = (event: LayoutChangeEvent) => {
+        setDimensions(event.nativeEvent.layout);
+    };
```
savelichalex: So, if a content is an Image we allocate unnecessary `state` and create onLayout callback, even though it's not used?

casts/cards/src/UICollectionCard/usePreload/usePreload.web.tsx
SAnatoliiS: The main purpose is not to add caching, but to preload data for whole list of images. Using this in UIImage won't achieve the goal.

casts/cards/src/UICollectionCard/CollectionSlide.tsx
SAnatoliiS: Fixed. Moved to separate component.

kit/media/src/UIVideo/UIVideo.web.tsx
SAnatoliiS: Both 😄

casts/cards/src/UICollectionCard/usePreload/usePreload.web.tsx
savelichalex: I meant not to implement this in UIImage but rather create a method `UIImage.prefetch` that hides platform dependent logic inside (because usage of RNImage on web, is kind of a detail that I don't want to leak outside) :)

savelichalex (approved):

kit/media/src/UIVideo/types.ts
```diff
@@ -28,17 +38,24 @@ export type UIVideoProps = {
...
+    resizeMode?: UIVideoResizeMode;
     /**
      * Callback function that is called when the media is loaded and ready to play
      */
     onLoad?: () => void;
     /**
      * Callback function that is called when video cannot be loaded
      */
-    onError?: () => void;
+    onError?: (error: any) => void;
```
SAnatoliiS, line 53: Here is `any` because type of error in `react-player` is `any` and IDK it's real format.

casts/cards/src/UICollectionCard/Preview.tsx
```diff
@@ -1,17 +1,23 @@
...
+export function Preview({ style, contentList }: PreviewProps) {
+    usePreload(contentList);
+
+    const currentSourceItemIndex = useCurrentSourceItemIndex(contentList);
+
+    const currentContent: MediaCardContent | null = React.useMemo(() => {
+        return contentList[currentSourceItemIndex];
+    }, [currentSourceItemIndex, contentList]);
+
+    if (contentList.length === 0 || !currentContent) {
```
sertony, line 27: Don't think that `contentList.length === 0` is still needed here, but it's OK.

casts/cards/src/UICollectionCard/useCurrentSourceItemIndex.ts
```diff
@@ -0,0 +1,38 @@
...
+
+function getShowTime(contentType: MediaCardContentType) {
+    switch (contentType) {
+        case 'Image':
+            return UIConstant.uiCollectionCard.timeToShowImage;
+        case 'Video':
+            return UIConstant.uiCollectionCard.timeToShowVideo;
+        case 'Unknown':
+        default:
+            return Infinity;
```
sertony: I'm scared to return `Infinity` here as L31 would give an unpredicted result and will allocate the timeout forever (keeping it memory).

I'd better suggest to return `undefined` and do not create the timeout with an Infinity

casts/cards/src/UICollectionCard/usePreload/usePreload.web.tsx
sertony: I actually agree with @savelichalex's statement

sertony: I actually agree with @savelichalex's statement

kit/media/src/UIVideo/types.ts
sertony: Hm... OK.

casts/cards/src/UICollectionCard/usePreload/usePreload.web.tsx
SAnatoliiS: Agree, done

sertony (approved):

PR #1386, opened 2021-12-03
add number type to badge

vetkamak:


casts/cards/src/UICollectionCard/types.ts
```diff
@@ -13,7 +13,7 @@ export type UICollectionCardProps = {
     /**
      * Displays the number of items in the collection or an image
      */
-    badge?: string | ImageSourcePropType;
+    badge?: string | number | ImageSourcePropType;
```
SAnatoliiS, line 16: Type `ImageSourcePropType` already contains type `number`. This is the type that the method `require` returns (at least on android). Consider to remove `number` here, so that there are no type collisions.

sertony: Hm, but actually I think that it's OK to pass the number (which is not a part of ImageSource).
In general I don't think it's gonna have any collisions with types here..
@savelichalex what do you think?

sertony (approved):

casts/cards/src/UICollectionCard/types.ts
SAnatoliiS: If you pass a picture to the component that you received via `require`, then this will break the component

SAnatoliiS: What's the problem with converting a number to a string to pass to badge?

savelichalex: I don't think there're any problems, but I do agree with @SAnatoliiS that it's better to pass badge number as a string :)

sertony: OK, hence this PR is not required ;) @vetkamak can you apply changes in Surf and close the PR?

PR #1387, opened 2021-12-06
Masonry list footer

SAnatoliiS:


kit/scrolls/src/Lists/Footer.tsx
```diff
@@ -0,0 +1,38 @@
...
+        if (ListFooterComponent) {
+            return React.isValidElement(ListFooterComponent) ? (
+                ListFooterComponent
+            ) : (
+                <ListFooterComponent />
+            );
+        }
+        return null;
+    }, [ListFooterComponent]);
+
```
savelichalex, line 22: Maybe return `null` if there isn't any ListFooterComponent provided?

kit/scrolls/src/Lists/UIMasonryList.tsx
```diff
@@ -236,6 +255,7 @@ function useVirtualization<Item>(
     const lastBottomCellIndex = React.useRef(0);
 
     const contentLengthOnEndReached = React.useRef(0);
+    const currentCoordinate = React.useRef(0);
```
savelichalex: Coordinate of what? Too abstract naming if you ask me :)

savelichalex: Btw, please add breadcrumbs why you have introduced such a variable)

kit/scrolls/src/Lists/UIMasonryList.tsx
```diff
@@ -255,7 +255,11 @@ function useVirtualization<Item>(
     const lastBottomCellIndex = React.useRef(0);
 
     const contentLengthOnEndReached = React.useRef(0);
-    const currentCoordinate = React.useRef(0);
+    /**
+     * Y value of the current scroll.
+     * We need to remember scroll value when the data has been updated.
```
savelichalex:
```suggestion
     * We need to remember scroll value 
     * to properly recalculate the virtual window
     * when a data has been updated.
```
am I right?

savelichalex (approved):

kit/scrolls/src/Lists/UIMasonryList.tsx
```diff
@@ -297,16 +323,34 @@ function useVirtualization<Item>(
...
         /**
          * So basically when we render items for the first time,
+         * or when the data has been updated,
          * we have to calculate items that're in virtual window
          */
         const bottom = binarySearch(
-            initialWindowHeight + virtualWindowSizeRatio * initialWindowHeight,
+            initialWindowHeight +
+                virtualWindowSizeRatio * initialWindowHeight +
+                currentScrollValue.current,
```
sertony, line 336: Not sure either it was a bug before? (when we haven't added the `currentScrollValue`)

SAnatoliiS: It was. When updating the data, the list disappeared (the viewport was calculated at the very beginning of the list)

kit/scrolls/src/Lists/UIMasonryList.tsx
```diff
@@ -297,16 +323,34 @@ function useVirtualization<Item>(
...
         );
-        manageBottomCells(lastBottomCellIndex.current, bottom, show, hide);
+        const top = binarySearch(
+            currentScrollValue.current -
+                virtualWindowSizeRatio * initialWindowHeight -
+                maxItemHeight,
+            cellsIndexes,
+        );
+
+        manageBottomCells(top, bottom, show, hide);
```
sertony: Not sure I understand the logic now.... Befoe we've compared current and prev bottom positions, now it's something new and the `top` item shouldn't be connected to the `bottom` management... Or I'm missing something. Can you clarify, please?

sertony (changes requested):
Please, read comments above

kit/scrolls/src/Lists/UIMasonryList.tsx
SAnatoliiS: There was a bug, when data was updated 
image

SAnatoliiS: After the data has been updated, some elements may appear above the last element of the previous list, so you need to update all the elements that fit on the screen and even little above.

savelichalex: I actually think that the whole effect isn't correct anymore. Before, as it was for first render only, I made an assumption that only bottom views should be managed (and actually this optimisation still can be applied if current scroll value is 0 or small enough (maybe less than virtual window?), but now we should do it "right" way:
```js
hide(0, top - 1)
show(top, bottom)
hide(bottom + 1, last)
```
PS. Not sure how to properly apply +-1, just because I don't remember whether the interval is inclusive or not :)

savelichalex (approved):

sertony (approved):

PR #1389, opened 2021-12-07
Lokalise: Translations update

localized26:
This pull request was initiated by Lokalise (user Khesha Tsyrenova) at 2021-12-07 08:18:29


sertony (changes requested):
@BigRoma can you help fixing the path to languages in lokalise.com?

PR #1390, opened 2021-12-07
add get image size failure handling

vetkamak:


sertony (approved):
Can you just leave a note about the failure callback? (that we want to use the maxImageSize and maybe why)

savelichalex (approved):

SAnatoliiS (approved):

PR #1391, opened 2021-12-07
Lokalise: Translations update

localized26:
This pull request was initiated by Lokalise (user Roman Tishakov) at 2021-12-07 12:38:59


sertony (changes requested):
It looks like we've added few more strings here which shouldn't be there. Can you check what's wrong?

localization/src/languages/ko.json
```diff
@@ -27,12 +27,12 @@
     "AllMembersCanAddAndRemoveMembers": "모든 구성원은 구성원을 추가 및 제거하고 그룹 이름과 사진을 편집할 수 있습니다.",
     "Amount": "액수",
     "Apartment": "Apartment/Suite/Unit",
-    "AtTime": "",
+    "AtTime": "at",
```
kkkhesha, line 30: @BigRoma в локалайзе не переведено сейчас, почему тут добавлен "at". Должно быть как в левой колонке - без перевода

PR #1394, opened 2021-12-08
Lokalise: Translations update

localized26:
This pull request was initiated by Lokalise (user Khesha Tsyrenova) at 2021-12-08 06:25:27


sertony:
Hm.. paths are wrong again. @BigRoma can you check that one more time, please?

PR #1403, opened 2021-12-16
Add UI image cell

SAnatoliiS:


savelichalex (commented):
We discussed it together with @sertony and we both agree, that there is no need in a separate package for this case, it can live in `uistory.browser` as it's needed only for Discover page. Also it can have less abstract naming.

savelichalex (approved):

stories/browser/src/constants.ts
```diff
@@ -0,0 +1,13 @@
+export const UIConstant = {
+    uiImageCell: {
```
sertony: rename it maybe? a.k.a `uiDebotCell`? :)

sertony (approved):

sertony (approved):

PR #1414, opened 2021-12-27
Add expandable text

SAnatoliiS:


SAnatoliiS:
image

localization/src/languages/ko.json
```diff
@@ -248,6 +248,9 @@
     "Exchange": "교환",
     "ExchangeFrom": "보낸 사람",
     "ExchangeTo": "받는 사람",
+    "ExpandableText": {
+        "more": "more..."
```
kkkhesha:
```suggestion
        "more": "더 보기"
```

kkkhesha: @SAnatoliiS

casts/texts/package.json
```diff
@@ -0,0 +1,87 @@
+{
+    "name": "@tonlabs/uicast.texts",
```
savelichalex, line 2: I'm not sure about `uicast.texts`, as UILabel IMO is a core component, and is tightly coupled with current themes (actually it's the only reason to use it over out-of-the-box Text from RN, that it provides correct font and  supports automatic theme switching).

@sertony what do you think?

SAnatoliiS: Components from `uicast.texts` (UIExpandableText) uses UILabel from `kit.themes`

sertony: @savelichalex I honestly don't understand what's your concern about. Is it about the cast name, or `UIExpandableText` component placement in this cast?

I got @SAnatoliiS 's idea about using `UILabel` from `kit.theme` and having this independent component build on top of it.

casts/texts/package.json
```diff
@@ -0,0 +1,87 @@
...
+        "react-native-gesture-handler": "^1.10.3",
+        "react-native-linear-gradient": "^2.5.6",
+        "react-native-reanimated": "2.3.1",
+        "typescript": "4.4.3"
+    },
+    "peerDependencies": {
+        "react": "^17.0.2",
+        "react-dom": "^17.0.2",
+        "react-native": "^0.65.0",
+        "react-native-fast-image": "^8.3.4",
```
sertony: Why having the dependency in this package?

casts/texts/package.json
```diff
@@ -0,0 +1,87 @@
...
+        "react-native-gesture-handler": "^1.10.3",
+        "react-native-linear-gradient": "^2.5.6",
```
sertony: Also, see no reason for that

casts/texts/src/UIExpandableText/UIExpandableText.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    const [isExpanded, setIsExpanded] = React.useState<boolean>(false);
+
+    const onExpand = React.useCallback(() => {
+        setIsExpanded(true);
+    }, []);
+
+    return (
+        <View style={styles.container} testID={testID}>
+            <UILabel {...props} numberOfLines={isExpanded ? undefined : numberOfLines} />
```
sertony: You well have two `testID` props in this render. One for <View /> and one for <UILabel />.
Consider changing to:
```suggestion
        <View style={styles.container} testID={testID}>
            <UILabel {...rest} numberOfLines={isExpanded ? undefined : numberOfLines} />
```
where `rest` is taken from
```
const { numberOfLines, testID, ...rest } = props;
```

sertony: Actually... do we need to mark View itself with the `testID`, but not a `label` itself?
If so, probably we'll also need to create an independent testID for UILabel, such as
```
testID={`UIExpandableText.Label:${testID}}
```

To let Detox find it.

casts/texts/package.json
savelichalex: I looked again, and now I'm OK with it :)

savelichalex (approved):

casts/texts/src/UIExpandableText/UIExpandableText.tsx
```diff
@@ -19,10 +19,14 @@ function UIExpandableTextImpl(props: UIExpandableTextProps) {
...
     return (
         <View style={styles.container} testID={testID}>
-            <UILabel {...props} numberOfLines={isExpanded ? undefined : numberOfLines} />
+            <UILabel
+                {...rest}
+                numberOfLines={isExpanded ? undefined : numberOfLines}
+                testID={`UIExpandableText.Text:${testID}`}
+            />
             {!isExpanded ? (
                 <TouchableOpacity onPress={onExpand} style={styles.moreButton}>
```
sertony: I guess we'll need to have a testID for "More" button as well

sertony (approved):

PR #1420, opened 2021-12-28
attempt to improve UICollectionCard perfomance

sertony:


casts/cards/src/UICollectionCard/useCurrentSourceItem.ts
```diff
@@ -0,0 +1,82 @@
...
+            if (index == null) {
+                // No items are available yet... Need to wait until they appear.
+                // Note: if items do not appear, they should automatically fill `failureIndexList`.
+                // Hence there is no need to wait if all of them fail to load!.
+                if (failureIndexList.current.length !== contentList?.length) {
+                    // Trigger the effect one more time in a sec to check the available list again
+                    timeout = setTimeout(() => setCurrentAvailableListIndex(0), 1000);
+                } else {
+                    // No need to iterate through failed items
+                }
```
sertony, line 51: Can't say I like this one, but it feels better than changing the state of the component when getting the available or failed item

savelichalex (approved):

SAnatoliiS (approved):

SAnatoliiS:
It doesn't display content on Android 
image
And content doesn 't switch in UICollectionCard on iOS and web

sertony:
> It doesn't display content on Android
> image
> And content doesn 't switch in UICollectionCard on iOS and web

That’s strange. It worked fine in Surf. I’ll check it in UIKit

casts/cards/src/UICollectionCard/useCurrentSourceItem.ts
```diff
@@ -0,0 +1,99 @@
...
+            if (index == null) {
+                // No items are available yet... Need to wait until they appear.
+                // Note: if items do not appear, they should automatically fill `failureIndexList`.
+                // Hence there is no need to wait if all of them fail to load!.
+                if (failureIndexList.current.length !== items.length) {
+                    // Trigger the effect one more time in a sec to check the available items again
+                    timeout = setTimeout(() => iterateThroughItems(items), 1000);
+                } else {
+                    // No need to iterate through failed items
+                }
```
sertony, line 58: Can't say I like this one, but it feels better than changing the state of the component when getting the available or failed item

sertony:
> It doesn't display content on Android
> image
> And content doesn 't switch in UICollectionCard on iOS and web

Wow, as for Android, I found out that FastImage's `onLoad` callback just doesn't trigger... 😕 

UPD. Found it, it happened to be that `display: none` removes the component from Android's DOM (unlike web & ios) completely 😮 

@SAnatoliiS @savelichalex FYI

casts/cards/src/UICollectionCard/CollectionSlide.tsx
```diff
@@ -73,7 +73,7 @@ function CollectionSlideImpl({ content, onLoad, onError, style, isVisible }: Col
     }, [content, contentType, onError]);
 
     const visibility = React.useMemo<StyleProp<ViewStyle>>(() => {
-        return { display: isVisible ? 'flex' : 'none' };
+        return { opacity: isVisible ? 1 : 0 };
```
savelichalex: Actually opacity is not the best solution to `hide` things, as on some platforms it disables optimisations. (Like from what I know how it works: usually during rendering, views layers (in terms of iOS) will be turned in textures, and then gpu will try to blend their pixels with regard of their positioning. So, from what I know, if a view is transparent, it's still going to take part in blending, that might load GPU with meaningless work. (more on that https://www.objc.io/issues/3-views/moving-pixels-onto-the-screen/).
So, we may do { top: isVisible ? 0 : -9999, left: isVisible ? 0 : -9999 }, that theoretically better for rendering.
Though I don't know how to measure it, it's just a theory :)

savelichalex: Btw, to save on GC calls you can do `isVisible ? styles.notTransparent : styles.transparent` :)

sertony: That’s actually true, thanks. My original idea was to make it invisible with display manipulation, but it happened to be not working in Android properly as the card was completely removed from DOM (unmounted), which was important for us to load the images while being invisible

SAnatoliiS:
One video in collection still doesn't work on Android. And catches many the same errors
image
image

savelichalex (approved):

SAnatoliiS (approved):

PR #1424, opened 2022-01-10
Improve go back action handling for split router

savelichalex:


casts/bars/src/useNavigationHeaderLeftItems.tsx
```diff
@@ -15,7 +15,16 @@ function findIfCanGoBackForStack(state: NavigationState, parentState?: Navigatio
     }
 
     if (state.type === 'split') {
-        const { nestedStack } = state as { nestedStack?: number[] };
+        const { nestedStack, isSplitted, history } = state as any as {
+            nestedStack?: number[];
+            isSplitted: boolean;
+            history: number[];
+        };
```
SAnatoliiS, line 22: Can we take out the typing separately? Inline typing makes the code harder to read and can lead to errors

SAnatoliiS (approved):

casts/bars/src/useNavigationHeaderLeftItems.tsx
savelichalex: It's actually because I had to cast it due to inability to use typings from split navigator, as it introduces cycles. So this type isn't going to be re-used and is going to be used in this particular place only.

sertony (approved):

PR #1429, opened 2022-01-12
UILargeTitleHeader refresh control improvements

savelichalex:


casts/bars/src/UILargeTitleHeader/LargeTitleHeaderContent.tsx
```diff
@@ -0,0 +1,160 @@
...
+                        shift.value,
+                        [0, rubberBandEffectDistance.value],
+                        [1, largeTitleScale.value],
+                        {
+                            extrapolateLeft: Extrapolate.CLAMP,
+                        },
+                    ),
+                },
+                {
+                    translateX: interpolate(
```
sertony, line 75: I wonder if we can "compensate" the scale in some other way without touching `translateX`...

It probably doesn't matter, because it was implemented in the same way before and worked great :)

casts/bars/src/UILargeTitleHeader/RefreshControl.tsx
```diff
@@ -1,73 +1,96 @@
...
-                refreshingGuard.value = false;
-                runOnJS(stopRefreshing)();
-            },
-        );
-    }, [onRefresh, forceChangePosition, refreshingGuard, stopRefreshing]);
+        try {
+            await onRefresh();
+        } catch (err) {
+            if (process.env.NODE_ENV === 'development') {
+                console.warn('Unhandled error catched during refresh:', err);
```
sertony:
```suggestion
                console.warn('Unhandled error has been caught during the refresh:', err);
```

casts/bars/src/UILargeTitleHeader/useRubberBandEffectDistance.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+// @inline
+const RUBBER_BAND_EFFECT_DISTANCE_WEB = 50;
+// @inline
+const RUBBER_BAND_EFFECT_DISTANCE_DEFAULT = 100;
+
+export const NON_UI_RUBBER_BAND_EFFECT_DISTANCE = Platform.select({
+    web: RUBBER_BAND_EFFECT_DISTANCE_WEB,
+    default: RUBBER_BAND_EFFECT_DISTANCE_DEFAULT,
+});
```
sertony, line 14: I wonder if it looks well on mobile web (probably it's not so important, but can be tested just in case)

savelichalex: I actually never did large title header properly for mobile web (didn't support touch events instead of wheel), so it might not work by design :D

sertony: OK :)

casts/bars/src/UILargeTitleHeader/LargeTitleHeaderContent.tsx
savelichalex: Maybe it's possible to use raw transformation matrix somehow, it's just I don't know how to use them yet 😅

sertony (approved):
This needs to be integrated in Surf :)

casts/bars/src/UILargeTitleHeader/LargeTitleHeaderContent.tsx
sertony: Also, it might be something connected to the view itself aligned to "center"... at least it would make sense

savelichalex: No, it's actually because scale is applied to a center of a view, there're APIs to specify scale origin (`transform-origin` on web), but it doesn't exist in RN

sertony (approved):

SAnatoliiS (approved):

PR #1433, opened 2022-01-14
Add ability to override whole navigation bar for UILargeTitleHeader

savelichalex:


casts/stackNavigator/src/StackNavigator/useWrapScreensWithUILargeTitleHeader.tsx
```diff
@@ -173,4 +178,11 @@ const styles = StyleSheet.create({
     screenContainer: {
         flex: 1,
     },
+    navBar: {
```
sertony, line 183: This is not used anywhere

sertony (approved):

PR #1436, opened 2022-01-14
UICollectionCardTestID

pheeel:


casts/cards/src/UICollectionCard/UICollectionCard.tsx
```diff
@@ -73,7 +73,9 @@ export function UICollectionCard({
                             </UILabel>
                         ) : null}
                     </View>
-                    <Badge badge={badge} />
+                    <View testID={`collection_badge_${title}${typeof badge === 'string' ? '_balance_' + badge :  ''}`}>
+                        <Badge badge={badge} />
+                    </View>
```
SAnatoliiS: I suggest moving the testID inside the Badge component so as not to change the layout and not to create additional nesting

SAnatoliiS: You may need to create a testID prop for the Badge if necessary

AleksandrHorev (approved):

casts/cards/src/UICollectionCard/UICollectionCard.tsx
```diff
@@ -73,7 +73,7 @@ export function UICollectionCard({
                             </UILabel>
                         ) : null}
                     </View>
-                    <Badge badge={badge} />
+                    <Badge badge={badge} testID={`collection_badge_${title}${typeof badge === 'string' ? '_balance_' + badge :  ''}`} />
```
sertony: I don't like seeing `balance` here... since it might be not balance, but smth else.
```suggestion
                    <Badge badge={badge} testID={`collection_badge_${title}:${badge.toString()}`} />
```

Why not like this?

SAnatoliiS (approved):

casts/cards/src/UICollectionCard/UICollectionCard.tsx
pheeel: If `badge` is a string - we render the balance of the collection, otherwise, we render the link icon, which I don't need here.

savelichalex: Maybe call it `count` then?

pheeel: Done @savelichalex

sertony (approved):

PR #1439, opened 2022-01-17
Add UIForeground.

SAnatoliiS:


SAnatoliiS:
image

savelichalex:
Is it just me, or `foreground` associates to a data fetching? :D

casts/rows/src/index.ts
```diff
@@ -1,6 +1,7 @@
 export * from './UIAccountRow';
 export * from './UILink';
 export * from './UICurrencyRow';
+export * from './UIForeground';
```
savelichalex: It shouldn't be in `rows`. `rows` package is for `rows` that is used in `VirtualizedList` and it's ancestors. Such component only will provide confusion when to use the package.

casts/rows/src/UIForeground/Parts/SecondaryPart.tsx
```diff
@@ -0,0 +1,34 @@
...
+        return {
+            disabled,
+            negative,
+            partType: 'Secondary',
+            partState: onPress || disabled !== undefined ? 'Pressable' : 'NonPressable',
+        };
+    }, [disabled, negative, onPress]);
+
+    return (
+        <TouchableOpacity
```
savelichalex: So it have unnecessary `TouchableOpacity` (with all it's complications) even if `onPress` won't be ever provided?

casts/rows/src/UIForeground/UIForeground.tsx
```diff
@@ -0,0 +1,36 @@
...
+    Container,
+
+    PrimaryPart,
+    SecondaryPart,
+
+    ActionElement,
+    IconElement,
+    NumberElement,
+    SectionElement,
+    TextElement,
```
savelichalex: Too verbose if you ask me, you can omit `Element` without loosing meaning (btw it contradicts with React terminology, where `element` means a component presented in JSX, aka the one that `React.createElement` returns)

casts/rows/src/UIForeground/UIForeground.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export function SectionElement() {
+    return <View />;
+}
+
+export const UIForeground: UIForegroundType = {
+    Container,
+
+    PrimaryPart,
+    SecondaryPart,
```
savelichalex: IMO `Part` is also unnecessary here

SAnatoliiS:
This is an internal component that will be used in many popups

casts/rows/src/index.ts
SAnatoliiS: Moved to kit.popups

casts/rows/src/UIForeground/UIForeground.tsx
SAnatoliiS: `Element` changed to `Cell`

casts/rows/src/UIForeground/UIForeground.tsx
SAnatoliiS: `Part` changed to `Column`

kit/popups/src/UIForeground/Container.tsx
```diff
@@ -0,0 +1,27 @@
...
+    return <View style={styles.container}>{children}</View>;
+}
+
+const styles = StyleSheet.create({
+    container: {
+        /**
+         * The negative right margin is used to set the right padding of all child elements
+         * and not to think about the presence of an element on the right side.
+         */
+        marginHorizontal: -UILayoutConstant.contentInsetVerticalX3 / 2,
```
savelichalex: I'm curious if a pressable area is correct on all platforms in that case

SAnatoliiS: I checked it in all platforms

savelichalex (approved):

sertony (approved):

PR #1441, opened 2022-01-18
Fix LocalizationService types.

BigRoma:


localization/src/service.ts
```diff
@@ -55,10 +55,9 @@ type LanguageServiceOptions<T> = {
 
 export type LocalizedStrings<T> = LocalizationServiceMethods & T;
 
-export type LocalizedInstance<T> = LocalizedStrings<T> & LocalizationService<T>;
+export type LocalizedInstance<T> = LocalizedStrings<T>;
```
sertony: Do we need it now as it's the same? :)

sertony (approved):

sertony (approved):

savelichalex (approved):

PR #1442, opened 2022-01-18
Make a constructor out of UISheet to support different behaviours

savelichalex:
Not every `Sheet` need the ability to handle a keyboard events (i.e. UIFullscreenSheet shouldn't adjust their position with keyboard opening), but it was built in core logic of the base `Sheet`, hence for UIFullscreen (and therefore for every modal in the app) it would have piece of code that just takes resources. So because of that I split implementation, and now `Sheet` can be composed from pieces that include only needed functionality.


kit/popups/src/Sheets/UISheet/UISheet.tsx
```diff
@@ -68,46 +54,49 @@ export type UISheetProps = {
      */
     hasCloseAnimation?: boolean;
     /**
-     * Whether UISheet should react on keyboard opening
-     *
-     * TODO(savelichalex): I was thinking maybe do it a bit smarter,
```
sertony, line 73: And you did it :D

sertony (approved):

SAnatoliiS (approved):

PR #1444, opened 2022-01-19
ui details test id fix

Khvoya:


packages/components/src/UIDetailsView/index.js
```diff
@@ -111,7 +111,6 @@ export default class UIDetailsView extends UIComponent<Props, State> {
         const testIDProp = testID ? { testID } : null;
         return (
             <Wrapper
-                {...testIDProp}
```
savelichalex, line 114: can you enlighten why? :)

packages/components/src/UIDetailsView/index.js
```diff
@@ -111,7 +111,6 @@ export default class UIDetailsView extends UIComponent<Props, State> {
         const testIDProp = testID ? { testID } : null;
```
sertony: Probably this also needs to be removed then :)

packages/components/src/UIDetailsView/index.js
Khvoya: test id duplicates

sertony (approved):

PR #1448, opened 2022-01-21
UIPageSheet

savelichalex:
A new sheet with fixed height and without a keyboard tracking (but maybe we need a helper for that, to calculate intersection between a sheet and a keyboard, or maybe a sheet can override insets with SafeAreaContext.Provider). Will be used primarily on desktops and tablets.


SAnatoliiS (approved):

kit/popups/src/Sheets/UIPageSheet.tsx
```diff
@@ -0,0 +1,69 @@
...
+    children,
+}: {
+    children: (
+        height: number,
+        getBottomInset: KeyboardAwareSheetProps['getBottomInset'],
+    ) => JSX.Element;
+}) {
+    const { height } = useWindowDimensions();
+
+    // TODO
```
sertony: TODO what?)

sertony (approved):

PR #1483, opened 2022-02-02
Add tooltip

SAnatoliiS:


kit/popups/src/useTargetDimensions.ts
```diff
@@ -0,0 +1,51 @@
...
+    const [targetDimensions, setTargetDimensions] = React.useState<TargetDimensions | null>(null);
+
+    React.useEffect(() => {
+        if (!targetRef.current || !targetRef.current.measure) {
+            console.error(
+                `[UITooltipContent]: [useTargetDimensions]: targetRef does not have a 'measure' method`,
+            );
+            return;
+        }
+        targetRef.current.measure(
```
savelichalex: Shouldn't it be `measureInWindow`? 🤔 or is it because of `Portal` that takes whole available space?

kit/popups/src/Tooltip/UITooltip.tsx
```diff
@@ -0,0 +1,71 @@
...
+     * Disabled `Pressable` is used here instead of `View` because the `onMeasure` method
+     * for the `ref` of the `View` does not return values on Android.
+     */
+    return (
+        <Pressable style={[styles.container, styleProp]} ref={containerRef} disabled>
+            {children}
+            {visible ? (
+                <UITooltipBox {...restProps} triggerRef={containerRef} onClose={onClose} />
+            ) : null}
+            <Pressable onPress={onOpen} style={styles.pressableOverlay} />
```
savelichalex: Can you please repeat why we need pressable overlay? (It would be cool to leave a comment too)

kit/popups/src/useTargetDimensions.ts
SAnatoliiS: I agree. I think it will be more correct to use `measureInWindow`

kit/popups/src/Tooltip/UITooltip.tsx
SAnatoliiS: I left a comment

SAnatoliiS: We have to position this pressable overlay in an absolute position due to the fact that we need to implement hitSlop by paddings, and the total size of the component should remain unchanged (the same size as children has)

savelichalex (approved):

kit/popups/src/Tooltip/UITooltip.tsx
SAnatoliiS: It doesn't seem to work for android. Then I will shift the need to configure hitSlop with paddings to users.

savelichalex (approved):

sertony (approved):

PR #1487, opened 2022-02-03
Fix UIAlertView with two actions.

SAnatoliiS:


savelichalex (approved):

sertony (approved):
Approved, but need to discuss with @savelichalex and Aleksey Matveev

PR #1501, opened 2022-02-14
fix UIMenuAction testID

Khvoya:


kit/popups/src/Menu/UIMenuAction.tsx
```diff
@@ -11,7 +11,7 @@ export const UIMenuAction: React.FC<UIMenuActionProps> = ({
         <UIForeground.Container key={title}>
             <UIForeground.PrimaryColumn>
                 <UIForeground.ActionCell
-                    testID={`${title}_menu_action_button`}
+                    testID={`${title}_action_button`}
```
sertony: What do you think about changing the name of the ID in your tests, instead of applying the change here? I think that `menu_action_button` is a better suffix than just `action_button`

savelichalex: it actually can be confused with UIActionButton I guess...

sertony (approved):

kit/popups/src/Menu/UIMenuAction.tsx
sertony: @Khvoya if you want to make it simple in test, how about changing it into?
```suggestion
                    testID={`${title}_action_button_menu`}
```

In that case you will only have to add `_menu` suffix to your constants on the desired screen

Khvoya: I did it because we have two similar modal menus with different item ids. For detox it's the same component and I led it to the single id

savelichalex: I understand nothing, but I guess that's why I always suggest to pass testIDs from user land

savelichalex (approved):
I will be honest, I don't like when testIDs changes because of Surf, not vice versa. The dependency chain is that Surf depends on UIKit, not UIKit on Surf

sertony:
> I will be honest, I don't like when testIDs changes because of Surf, not vice versa. The dependency chain is that Surf depends on UIKit, not UIKit on Surf

How about providing a `testID` property for `Action` components instead?
@SAnatoliiS can you do it?

savelichalex (approved):
Thanks @SAnatoliiS !

sertony (approved):

PR #1505, opened 2022-02-17
Update material text view

SAnatoliiS:


savelichalex (approved):

kit/inputs/src/UIMaterialTextView/UIMaterialTextViewFloating.tsx
```diff
@@ -0,0 +1,179 @@
...
+import { useAutogrow } from './useAutogrow';
+import { UIMaterialTextViewComment } from './UIMaterialTextViewComment';
+import { useExpandingValue } from './useExpandingValue';
+
+// @inline
+const POSITION_FOLDED: number = 0;
+// @inline
+const POSITION_EXPANDED: number = 1;
+
+// EXPANDED_INPUT_OFFSET = UILayoutConstant.contentOffset - UILayoutConstant.contentInsetVerticalX2
```
sertony, line 24: What about this comment? Maybe kill it if it's not needed?

kit/inputs/src/UIMaterialTextView/useClearButton.tsx
```diff
@@ -22,9 +25,11 @@ export function useClearButton(
                     onPress={clear}
```
sertony, line 25: @SAnatoliiS btw, have you found out why this button wasn't pressing on Android device?

sertony (approved):
Approved, but we won't merge it until @vetkamak integrates it in Surf ;)

savelichalex:
> Approved, but we won't merge it until @vetkamak integrates it in Surf ;)

I thought it doesn't have breaking changes. Correct me if I'm wrong @SAnatoliiS

SAnatoliiS:
@savelichalex is right. There is no breaking changes, only additional prop - `warning?: boolean`.

kit/inputs/src/UIMaterialTextView/UIMaterialTextViewFloating.tsx
SAnatoliiS: It says here how this constant is calculated

kit/inputs/src/UIMaterialTextView/useClearButton.tsx
SAnatoliiS: Yes, I fixed it.

sertony:
> @savelichalex is right. There is no breaking changes, only additional prop - `warning?: boolean`.

Do we need to apply it anywhere in Surf?
Also about the missing `floating` prop, what component should we use instead when we want to "hide" the label, but show it as the placeholder (as it's done in our current InvitationView in Surf)

SAnatoliiS:
Yes, I forgot about prop `floating`. We need to remove it from surf. Do not mix `Label` and `Placeholder`. If we need to hide Label, we don't need to pass any string to it (I made it optional). `floating={false}` for now is equal to `label=""` or missing `label` prop.

savelichalex:
> Yes, I forgot about prop `floating`. We need to remove it from surf. Do not mix `Label` and `Placeholder`. If we need to hide Label, we don't need to pass any string to it (I made it optional). `floating={false}` for now is equal to `label=""` or missing `label` prop.

woah, I actually like it, that makes a lot of sense! 👍

SAnatoliiS:
I made integration PR in Surf https://github.com/tonlabs/TON-Surf/pull/3201

PR #1515, opened 2022-02-28
Fix text align

BigRoma:


stories/browser/src/UICellDebot/UICellDebot.tsx
```diff
@@ -30,26 +30,26 @@ function UICellDebotImpl({ image, title, caption, onPress, loading, testID }: UI
...
-                            {caption}
-                        </UILabel>
-                    ) : null}
-                </View>
+                {title ? (
+                    <UILabel
+                        role={TypographyVariants.NarrowActionFootnote}
+                        color={ColorVariants.TextPrimary}
+                        numberOfLines={caption ? 1 : 2}
+                        style={styles.textContainer}
```
sertony, line 38: UILabel does not have `style` prop (and can be used only for internal purposes).
@savelichalex any ideas?

savelichalex: Let's say it's still internal, as it's in @story :)

stories/browser/src/UICellDebot/UICellDebot.tsx
```diff
@@ -30,26 +30,26 @@ function UICellDebotImpl({ image, title, caption, onPress, loading, testID }: UI
...
+                    >
+                        {title}
+                    </UILabel>
+                ) : null}
+                {caption ? (
+                    <UILabel
+                        role={TypographyVariants.NarrowParagraphFootnote}
+                        color={ColorVariants.TextSecondary}
+                        numberOfLines={title ? 1 : 2}
+                        style={styles.textContainers}
```
savelichalex: There is no `textContainers` is it a miss-type?

stories/browser/src/UICellDebot/UICellDebot.tsx
sertony: What about TypeScript?)

savelichalex: @BigRoma does it pass TS checks?

sertony (approved):

PR #1518, opened 2022-03-01
Currency updates

vetkamak:


stories/chats/src/UIChatInput/styles.tsx
```diff
@@ -18,4 +18,8 @@ export const commonStyles = StyleSheet.create({
         height: UIConstant.mediumIconSize(),
         width: UIConstant.mediumIconSize(),
     },
+    actionTitle: {
+        minWidth: UIConstant.iconSize(),
```
savelichalex: What's the reason to set minWidth? can you a comment here to not forget what it's for?

vetkamak: I'll make an update and remove it, it won't be needed anymore))

sertony (approved):

savelichalex (approved):

PR #1519, opened 2022-03-01
Add amount input

SAnatoliiS:


kit/inputs/src/useApplyMask/amount/onChangeAmount.ts
```diff
@@ -0,0 +1,38 @@
...
+    moveCarret: MoveCarret,
+    skipNextOnSelectionChange: SharedValue<boolean>,
+) {
+    const {
+        grouping: integerSeparator,
+        decimal: delimeter,
+        decimalGrouping: fractionalSeparator,
+    } = uiLocalized.localeInfo.numbers;
+
+    const inputSeparators = [',', '.', '/', 'б', 'ю'];
```
sertony: this Cyrillic symbols scares me... (I don't like it here for sure!)
You can take it from the `delimeter` above...

kit/inputs/src/useApplyMask/amount/runUIGetSeparatorRegExp.ts
```diff
@@ -0,0 +1,27 @@
...
+        if (separator === integerSeparator || separator === fractionalSeparator) {
+            /** The currently used separators should not be used as delimeter */
+            return result;
+        }
+        // TODO Add more specific chars
+        if (separator === '.' || separator === '/') {
+            const separatorWithEscapeCharacter = `\\${separator}`;
+            result.push(separatorWithEscapeCharacter);
+            return result;
+        }
```
sertony: This also scares me)

kit/inputs/src/useApplyMask/useApplyMask.tsx
```diff
@@ -0,0 +1,54 @@
...
+        useOnSelectionChange();
+
+    const lastNormalizedText = useSharedValue('');
+    const lastText = useSharedValue('');
+
+    const onChangeText = React.useCallback(
+        (text: string): void => {
+            switch (mask) {
+                case '[000] [000].[000] [000]':
+                    onChangeAmount(
```
sertony: So, it's basically just a custom name now)

kit/inputs/src/useApplyMask/amount/onChangeAmount.ts
SAnatoliiS: I've added new `decimalAlternative` to `uiLocalized.localeInfo.numbers`.

kit/inputs/src/useApplyMask/amount/runUIGetSeparatorRegExp.ts
SAnatoliiS: We need to escape specific characters. I added full list of specific characters on top of the file.

kit/inputs/src/useApplyMask/useApplyMask.tsx
SAnatoliiS: This is a temporary solution so as not to complicate this task

kit/inputs/src/useApplyMask/amount/runUIGetDelimiterRegExp.ts
```diff
@@ -0,0 +1,32 @@
+const specialCharacters = '[]\\^$.|?*+()';
```
savelichalex: It's actually not a good idea to keep it outside of the function, as it will create a shared value for this constant and will keep it in memory all the time, while when it's created in function it will be created only on UI JS runtime and will be GCed after the function invocation.
(Exactly why I introduced numbers inlining)

kit/inputs/src/useApplyMask/amount/onChangeAmount.ts
```diff
@@ -0,0 +1,37 @@
...
+        rawNumber,
+        selectionStart,
+        selectionEnd,
+        integerSeparator,
+        delimeter,
+        fractionalSeparator,
+        lastNormalizedText,
+        lastText,
+        changeText,
+        moveCarret,
```
savelichalex: Actually to pass functions that're not a worklet might be not the best idea. Can we instead return sth and then apply it?

kit/inputs/src/useApplyMask/amount/onChangeAmount.ts
```diff
@@ -0,0 +1,37 @@
...
+    skipNextOnSelectionChange: SharedValue<boolean>,
+) {
+    const {
+        grouping: integerSeparator,
+        decimal: delimeter,
+        decimalGrouping: fractionalSeparator,
+        decimalAlternative: delimeterAlternative,
+    } = uiLocalized.localeInfo.numbers;
+
+    runOnUI(runUIOnChangeAmount)(
```
savelichalex: I'm starting to think, that as for now we don't catch events in native (hence don't run worklets directly), then maybe for now we don't have to call `runOnUI`, it won't make anything better, but rather kinda harmful as it involve few context switches and a lot of `SharedValue`s, that increase memory consumption.

It's cool that you did it with worklets, and I hope soon we'll get benefits from them, but for now I propose to call them directly, without threads switching.

kit/inputs/src/useApplyMask/amount/runUIFormat.ts
```diff
@@ -0,0 +1,71 @@
...
+        integerSeparator,
+    );
+
+    result.push(groupedIntegerPart);
+
+    // Normalize and group fractional part
+    if (DECIMALS > 0 && decimalPart != null) {
+        let normalizedDecimalPart = decimalPart.replace(notNumbersRegexp, '');
+
+        normalizedDecimalPart = normalizedDecimalPart.slice(0, DECIMALS);
```
savelichalex, line 53: hm, why it's restricted to 8 characters? I thought it's a general purpose method

kit/inputs/src/useApplyMask/amount/runUIFormat.ts
```diff
@@ -0,0 +1,71 @@
...
+    'worklet';
+
+    const inputDelimeterRegExp = runUIGetDelimiterRegExp(
+        integerSeparator,
+        fractionalSeparator,
+        delimeterAlternative,
+    );
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [integerPart, _inputDelimeter, decimalPart] = rawNumber.split(inputDelimeterRegExp);
```
savelichalex: I can't say that I understand what is going on here :)

Am I right that for now it works by coincidence? It just happened that `inputDelimeterRegExp` return `/(,)/` (or whatever delimiter for now is), but for some reason it accepts an array of delimiters, then I'm not sure how it suppose to work.

So I did a quick test.
I ran `runUIGetDelimiterRegExp(' ', ' ', ['a', 'b'])` and got `/(A|a|B|b)/`. 
That's four delimiters.
Then I created a (weird) string that follows this pattern - `'123A456a789b000'` and ran `'123A456a789b000'.split(/(A|a|B|b)/)`. 
It returns `['123', 'A', '456', 'a', '789', 'b', '000']`. 
So in this code we have integerPart - 123, decimalPart - 456 and the rest is just gone.

Seems overcomplicated IMO.

P.S. It would be great if you created unit tests for this.

kit/inputs/src/useApplyMask/useApplyMask.tsx
```diff
@@ -0,0 +1,54 @@
...
+                        selectionEnd,
+                        lastNormalizedText,
+                        lastText,
+                        changeText,
+                        moveCarret,
+                        skipNextOnSelectionChange,
+                    );
+                    break;
+                default:
+                    changeText(text);
```
savelichalex: Is there any reason to call it instead of doing nothing except to call a `onChangeTextProp`? (I just don't see a reason why force change a value that already on place)

savelichalex: PS. And yes, I know, that `changeText` will call `onChangeTextProp`, but it also has a `setNativeProps` call as well, and that what bother me

kit/inputs/src/useApplyMask/amount/runUIFormat.ts
savelichalex: Ok, I guess I got your assumption, that at any time there will be only one of such delimiters on a text, and it will be immediately replaced to the proper one. But the question is that assumption correct? 🤔

localization/android/src/main/java/tonlabs/uikit/localization/UIKitLocalizationModule.java
```diff
@@ -41,6 +41,16 @@ public String getName() {
...
+        numbers.put("decimalAlternative", new String[] {
+                "\u044E",
+                "\u0431",
+                "/",
+                "?",
+                "<",
+                ">",
+                ",",
+                "."
+        });
```
savelichalex, line 53: Can we say that it's safe to replace them on all keyboard layouts? 🤔

savelichalex: Seems to me that it will properly work only on Russian layout :) Maybe I'm wrong

localization/src/utils/getters.ts
```diff
@@ -11,6 +11,7 @@ export function getNumberFormatInfo(): NumberFormatInfo {
         thousands: parts[2],
         decimal: parts[3],
         decimalGrouping: '\u2009',
+        decimalAlternative: ['\u044E', '\u0431', '/', '?', '<', '>', ',', '.'],
```
savelichalex, line 19: What is the reasoning to put in UILocalized if they're actually constants? It doesn't depend on user language. @sertony what do you think?

kit/inputs/src/UIMaterialTextView/types.ts
```diff
@@ -1,19 +1,88 @@
+import type React from 'react';
 import type { View, TextInput } from 'react-native';
+import type { UIImageProps } from '@tonlabs/uikit.media';
 import type { UITextViewProps } from '../UITextView';
 import type { OnHeightChange } from '../useAutogrowTextView';
 
-export type UIMaterialTextViewProps = UITextViewProps & {
+export type UIMaterialTextViewMask = '[000] [000].[000] [000]';
```
savelichalex: What do you think to just call it `amount` or `@amount` (to emphasize that it's custom) as for now it might just confuse everyone.

savelichalex: (Or we can look at the `type` attribute for inputs in HTML as a reference)

kit/inputs/src/UIMaterialTextView/UIMaterialTextView.tsx
```diff
@@ -1,20 +1,116 @@
...
+        onHeightChange,
+    );
+
+    const { changeText, moveCarret } = useChangeImperative(
+        ref,
+        multiline,
+        onChangeTextWithInputHasValue,
+    );
+
+    const { onChangeText, onSelectionChange } = useApplyMask(changeText, moveCarret, mask);
```
savelichalex: I can't say that I like it to be applied to **every** input, taking into account the fact, that actually there will be less inputs with mask than without it, so it means that most of the inputs will do meaningless job (and will just waste CPU resources).

kit/inputs/src/UIMaterialTextView/hooks/useChangeImperative.ts
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
+import { Platform, TextInput } from 'react-native';
+import { calculateWebInputHeight } from '../../useAutogrowTextView';
+import { moveCarret as moveCarretPlatform } from '../../moveCarret';
+import type { ChangeText } from '../types';
+
+export function useChangeImperative(
```
savelichalex: What do you say to call it sth like `useChangeValueDirectly`, I personally didn't understand what `imperative` means in this context :)

savelichalex: Oh, I got it, it's a reference to `useImperativeHandle`! then maybe `useImperativeChange`? :)

kit/inputs/src/useApplyMask/amount/onChangeAmount.ts
savelichalex: Like
```ts
const { text, selection } = runOnUIChangeAmount(...);

changeText(text);
moveCarret(selection);
```

kit/inputs/src/UIMaterialTextView/types.ts
```diff
@@ -1,19 +1,88 @@
...
+    /**
+     * Label of the UIMaterialTextView
+     */
     label?: string;
+    /**
+     * Text providing validation information
+     */
     helperText?: string;
+    /**
+     * Is validation error?
```
savelichalex: If it was a question to me then - yes :)

localization/src/utils/getters.ts
savelichalex: Discussed in PM:
- Probably it worse to leave a TODO comment to "think about language-based values for this purpose" and leave a descriptive comment why it's needed (in order to not forget why you introduced it after all)

kit/inputs/src/useApplyMask/amount/runUIFormat.ts
savelichalex: Discussed in DM:
1. Leave a comment that we have an assumptions that acceptable string can have only 1 delimiter and that it's ok to remove parts from intentionally invalid string that can be copy-pasted.

kit/inputs/src/useApplyMask/amount/runUIFormat.ts
SAnatoliiS: Made 9, as it was in UITransferInput.

kit/inputs/src/useApplyMask/amount/runUIFormat.ts
SAnatoliiS: I returned the previous behavior: the entire part after the first separator is filtered and taken as the decimal part .

kit/inputs/src/useApplyMask/useApplyMask.tsx
SAnatoliiS: I made tweak of the `changeText` function. For now it takes config object as a second parameter. This config has `shouldSetNativeProps` field for not to call `setNativeProps`.
Note: We do need to call `setNativeProps` if clear button was tapped and we need to clean input.

kit/inputs/src/UIMaterialTextView/types.ts
SAnatoliiS: We discussed this problem and decided to use reserved words for strings whose length is not limited. For example, I called this formatting `Amount`.

kit/inputs/src/UIMaterialTextView/UIMaterialTextView.tsx
SAnatoliiS: As we discussed, I put the `mask` in the `ref` and called further hooks by condition, the result of which cannot be changed

kit/inputs/src/UIMaterialTextView/types.ts
SAnatoliiS: Rephrased the comment

kit/inputs/src/useApplyMask/amount/onChangeAmount.ts
```diff
@@ -0,0 +1,47 @@
...
+    skipNextOnSelectionChange: SharedValue<boolean>,
+) {
+    const {
+        grouping: integerSeparator,
+        decimal: delimeter,
+        decimalGrouping: fractionalSeparator,
+        decimalAlternative: delimeterAlternative,
+    } = uiLocalized.localeInfo.numbers;
+
+    const { formattedText, normalizedText, carretPosition } = runUIOnChangeAmount(
```
savelichalex, line 23: 👍

savelichalex (approved):

kit/inputs/src/useApplyMask/useApplyMask.tsx
```diff
@@ -0,0 +1,64 @@
...
+) {
+    /**
+     * The prop `mask` must not be changed, so we can use ref to make sure it will not.
+     * NOTE: Do not reassign this ref!
+     */
+    const maskRef = React.useRef(mask);
+
+    React.useEffect(() => {
+        if (maskRef.current !== mask) {
+            console.error(`[UIMaterialTextView] useApplyMask.ts: prop 'mask' must not be changed`);
```
sertony, line 48: 👍

sertony (approved):

PR #1525, opened 2022-03-03
Fix UIFullscreenSheet web height

savelichalex:


kit/popups/src/Sheets/UIFullscreenSheet.tsx
```diff
@@ -11,13 +12,17 @@ export type UIFullscreenSheetProps = UISheetProps & {
...
     } = useDimensions();
 
-    const fullscreenHeight = React.useMemo(
-        () => height + UILayoutConstant.rubberBandEffectDistance,
-        [height],
-    );
+    const fullscreenHeight = React.useMemo(() => {
+        if (Platform.OS === 'web') {
+            return windowHeight + UILayoutConstant.rubberBandEffectDistance;
+        }
```
sertony, line 22: Why should it be applied only for web?
Why not using `windowHeight` for all platforms here?

sertony (approved):

sertony (approved):

PR #1526, opened 2022-03-03
Add forId prop into UISeedPhraseTextView.

SAnatoliiS:


savelichalex (approved):

package.json
```diff
@@ -65,7 +65,7 @@
         "html2canvas": "1.0.0-rc.7",
         "mobile-detect": "1.4.3",
         "nanoid": "3.1.31",
-        "qrcode": "1.4.4",
+        "qrcode": "^1.4.4",
```
sertony, line 68: Why so? This will install `1.5.0` version in Surf (while keeping @types/qrcode: "1.4.0")

sertony (approved):

PR #1528, opened 2022-03-03
Fix android multiline material input

SAnatoliiS:


savelichalex (commented):
Are you sure about not using `useAutogrowTextView` for mobile platforms? Have you checked that on all platforms? Have you checked that in chat and browser?

sertony:
> Are you sure about not using `useAutogrowTextView` for mobile platforms? Have you checked that on all platforms? Have you checked that in chat and browser?

yep, that also doesn't feel right to me...

SAnatoliiS:
I have checked on all platforms (android/ios/web) in inputs, chat and browser.

SAnatoliiS:
TextInput needs nothing to change it's height in multiline mode. But using `useAutogrow` leads to reset input value when word doesn't fits in line.

savelichalex:
What about constrainedHeight? Does it still work?

kit/inputs/src/UIMaterialTextView/useAutogrow.ts
```diff
@@ -1,88 +1,21 @@
...
 export function useAutogrow(
-    ref: React.Ref<TextInput>,
+    _ref: React.Ref<TextInput>,
     onContentSizeChangeProp: UIMaterialTextViewProps['onContentSizeChange'],
     onChangeProp: UIMaterialTextViewProps['onChange'],
-    multiline: UIMaterialTextViewProps['multiline'],
+    _multiline: UIMaterialTextViewProps['multiline'],
     numberOfLinesProp: UIMaterialTextViewProps['numberOfLines'],
-    onHeightChange?: OnHeightChange,
+    _onHeightChange?: OnHeightChange,
```
sertony, line 12: I see this is used in so many places, hence it needs to be fixed differently...

PR #1531, opened 2022-03-05
RN 0.67.3

savelichalex:


Example/android/build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
         minSdkVersion = 21
         compileSdkVersion = 30
         targetSdkVersion = 30
-        ndkVersion = "20.1.5948944"
+        ndkVersion = "21.4.7075529"
```
sertony, line 9: https://facebook.github.io/react-native/docs/building-from-source#download-links-for-android-ndk - for some reason they haven't updated the download link to the newer version

Example/package.json
```diff
@@ -67,12 +67,11 @@
         "qrcode": "1.4.4",
         "react": "17.0.2",
         "react-dom": "17.0.2",
-        "react-native": "0.65.1",
-        "react-native-animateable-text": "0.7.0",
+        "react-native": "0.67.3",
```
sertony, line 68: What about `metro-react-native-babel-preset`?

Example/reinstall.sh
```diff
@@ -1,9 +1,12 @@
 npx lerna bootstrap
 npx lerna run prepare
 
+rvm use ruby-$(cat .ruby-version)
+bundle install
+
 cd ios
 rm -rf Pods
-pod install
+bundle exec pod install
```
sertony, line 9: oh :)

casts/dateTimePicker/src/UIDateTimePicker/Calendar/Years.tsx
```diff
@@ -41,10 +41,10 @@ function getYearLabelColor(disabled: boolean, isSelected: boolean) {
 export const Years = React.memo(function Years() {
     const { yearsMatrix, selectedRow, selectedYear, minYear, maxYear, onSelect } = useYears();
     const theme = useTheme();
-    const listRef = React.useRef<FlatListType>(null);
+    const listRef = React.useRef<any>(null);
```
sertony, line 44: Probably it just wanted to see `<RNFlatList | null>`?

index.js
```diff
@@ -1,9 +0,0 @@
-/**
- * @format
- */
-
-import { AppRegistry } from 'react-native';
-import { name as appName } from './Example/app.json';
-import App from './Example/src/App';
-
-AppRegistry.registerComponent(appName, () => App);
```
sertony, line 9: 😅

kit/themes/src/AnimateableText/AnimateableText.web.tsx
```diff
@@ -39,4 +39,4 @@ class AnimateableTextWeb extends React.Component<Text['props'] & { text: string
     }
 }
 
-export const AnimateableText = Animated.createAnimatedComponent(AnimateableTextWeb);
+export const AnimateableText: any = Animated.createAnimatedComponent(AnimateableTextWeb);
```
sertony, line 42: What was the problem?

postinstall.sh
```diff
@@ -6,10 +6,6 @@ sed -i '' -e '195s/.*/      } else if (typeof strings[key] !== "string" \&\& typ
 # Patch Android
 ## Remove `jcenter` in favour of `google` or `mavenCentral`
 
-### Patch `react-native-animateable-text`
-sed -i '' -e '7s/.*/    mavenCentral()/' ./node_modules/react-native-animateable-text/android/build.gradle
-sed -i '' -e '51s/.*//' ./node_modules/react-native-animateable-text/android/build.gradle
-
```
sertony, line 12: 👍

casts/dateTimePicker/src/UIDateTimePicker/Calendar/Years.tsx
savelichalex: nah, it's sth completely wrong with types of scrollables that I export in `uikit.scrolls`, for now I just don't understand how it can be different regarding that I copy-pasted it from RN types and simply should be the same :)

savelichalex: I actually had the same problem in UIChatList, didn't find how to fix it either

Example/package.json
savelichalex: Thanks for highlighting, I checked and it's already `0.66.2`, almost the same as RN updater suggest: `"metro-react-native-babel-preset": "^0.66.2",`, though I see on npm that it already has [`0.69.0` version](https://www.npmjs.com/package/metro-react-native-babel-preset) (maybe that's why better to use a cap `^` 🤔 )

Example/android/build.gradle
savelichalex: yep, also saw it :) but I actually was surprised how smooth my upgrade to the new NDK was, not as painful as before :)

index.js
savelichalex: btw, do you know what the heck was it? :D

kit/themes/src/AnimateableText/AnimateableText.web.tsx
savelichalex: Very weird TS error https://github.com/microsoft/TypeScript/issues/42873, I tried few things from the thread, but it didn't help, what's surprising is that `@ts-ingore` didn't work either (it's first time when I can't suppress an error). Since it has `.web` extension and the actual type is get from the file without extension it won't break anything, so I decided just suppress the error with `any` and don't mess with `tsconfig`

casts/dateTimePicker/src/UIDateTimePicker/Calendar/Years.tsx
sertony: OK)

index.js
sertony: I guess you've used it in order to run RN when "main" property of package.json was equal to "index.js" =)

kit/themes/src/AnimateableText/AnimateableText.web.tsx
sertony: But don't you loose an autocomplete here when developing for this component?

savelichalex: Actual type goes from https://github.com/tonlabs/UIKit/blob/development/kit/themes/src/AnimateableText/AnimateableText.tsx#L3 , since TS doesn't understand suffixes on files :)

index.js
savelichalex: I hope I wasn't drunk when I did it, it just doesn't make sense for me now :D

SAnatoliiS (approved):

PR #1533, opened 2022-03-09
Fix UISeedPhrasePopover location on android.

SAnatoliiS:


savelichalex:
It should point to surf@7.1.0 or it would break on modals integration (I assume)

sertony (approved):

kit/inputs/src/UISeedPhraseTextView/UISeedPhrasePopover.tsx
```diff
@@ -21,8 +13,11 @@ import { Portal, UILayoutConstant } from '@tonlabs/uikit.layout';
 
 const MAX_CELLS = 3;
 
+const statusBarHeight =
+    Platform.OS === 'android' && StatusBar.currentHeight ? StatusBar.currentHeight : 0;
```
savelichalex: I don't think that you should check for platform, as this constant works only for Android anyway

SAnatoliiS: This is a protection against possible future changes if this constant returns something for ios as well. I think it won't be superfluous.

savelichalex: It actually never worked for iOS :) Let's not make it overcomplicated, if it will be changed it will break a lot of places not only this one :)

SAnatoliiS: Done

savelichalex (approved):

PR #1536, opened 2022-03-16
Update notice colors

SAnatoliiS:


kit/popups/src/Notice/toastNoticeHooks/useTopToastNoticeYSnapPoints.ts
```diff
@@ -1,9 +1,9 @@
 import * as React from 'react';
 import { useWindowDimensions } from 'react-native';
 import Animated, { useDerivedValue } from 'react-native-reanimated';
-import { UIDevice } from '@tonlabs/uikit.core';
-import { UILayoutConstant } from '@tonlabs/uikit.layout';
 import { initialWindowMetrics } from 'react-native-safe-area-context';
+import { getStatusBarHeight } from 'react-native-iphone-x-helper';
```
sertony: AFAIK, @savelichalex wanted to remove this dependency from the project (and actually did it in https://github.com/tonlabs/UIKit/pull/1531/). Can we measure the status bar differently?

savelichalex: Actually better way would be to use top inset from safe area, just like it's done for UIModalSheet

savelichalex (commented):
Agree with @sertony that `react-native-iphone-x-helper` isn't needed, other libs that used it already got rid of it.

sertony (approved):

savelichalex (approved):

PR #1540, opened 2022-03-17
Fix multiline unput

SAnatoliiS:


kit/inputs/src/UIMaterialTextView/useAutogrow.ts
```diff
@@ -1,88 +1,45 @@
...
 
     const onContentSizeChange = React.useCallback(
         (event: any) => {
-            if (onAutogrowContentSizeChange) {
-                onAutogrowContentSizeChange(event);
+            if (onHeightChange) {
+                const height = event?.nativeEvent?.contentSize?.height;
+                if (height !== inputHeight.current) {
+                    inputHeight.current = height;
+                    if (typeof height === 'number') {
```
savelichalex: Could you please put a comment why you had to check for type of a number, as it's unclear what was the problem

SAnatoliiS: I put down the type of the `event` and realized that the check was unnecessary. Removed

savelichalex (approved):
I'm still a bit cautious about all this removes of logic, but I can't remember why exactly that was necessary and frankly it all looks like not a properly described set of hacks, so I'm OK with this changes, let's try it at least :)
But we for sure need to sit down on all those hacks and look if we can turn them to sth more stable.

sertony (approved):

PR #1547, opened 2022-03-23
Add number UI amount input

SAnatoliiS:


packages/components/src/index.js
```diff
@@ -1,10 +1,7 @@
 // @flow
-export { default as UIAmountInput } from './UIAmountInput';
 export * from './UIActionComponent';
 export { default as UIComponent } from './UIComponent';
 export { default as UIDetailsTable } from './UIDetailsTable';
 export { default as UIDetailsView } from './UIDetailsView';
-export * from './UIDetailsInput';
 export { default as UIPureComponent } from './UIPureComponent';
 export { default as UIScaleButton } from './UIScaleButton';
```
sertony, line 6: Can you make a new component for that in UIKit, instead of this one, please? (@savelichalex can you manage it 🙏 )
It's a small one, but would let us kill `UIActionComponent` and `UIPureComponent` at once :)

sertony (approved):
Does it have any breaking changes?

packages/components/src/index.js
savelichalex: I'm actually curious what it's used for?

sertony: Right now this kind of button is used for PromoCards in Surf

sertony: oh, and also for transactions bubbles in the chat ;)

savelichalex: aha, so it's not an actual button (with it's own design), it's kinda like `TouchableScale`

sertony: True!

SAnatoliiS:
No, it doesn't have any breaking changes

kit/inputs/src/UIAmountInput/hooks.ts
```diff
@@ -0,0 +1,119 @@
...
+export function useOnChangeText(
+    onChangeAmount: (amount: BigNumber | undefined) => void,
+    ref: React.RefObject<UIMaterialTextViewRef>,
+) {
+    const previousTextRef = React.useRef<string>('');
+    return React.useCallback(
+        (text: string): void => {
+            const amount = convertTextToBigNumber(text);
+            if (amount?.isNaN()) {
+                console.error(
```
savelichalex, line 40: I'm curious whether it will spam a lot in console in production or not

kit/inputs/src/UIAmountInput/hooks.ts
```diff
@@ -0,0 +1,119 @@
...
+                return;
+            }
+            onChangeAmount(amount);
+            previousTextRef.current = text;
+        },
+        [onChangeAmount, ref],
+    );
+}
+
+export function useHelperTextStatus(messageType: UIAmountInputMessageType | undefined) {
```
savelichalex, line 53: omg 😅

kit/inputs/src/UIAmountInput/hooks.ts
```diff
@@ -0,0 +1,119 @@
...
+        function changeAmount(amount: BigNumber | undefined, callOnChangeProp?: boolean) {
+            localRef.current?.changeText(amount ? amount.toString() : '', callOnChangeProp);
+        },
+        [localRef],
+    );
+
+    React.useImperativeHandle<UIMaterialTextViewRef, UIAmountInputRef>(
+        forwardedRed,
+        (): UIAmountInputRef => ({
+            ...(localRef.current ? localRef.current : UIMaterialTextView.prototype),
```
savelichalex: I doubt that `UIMaterialTextView.prototype` will work as it's not a class, it's a function :) so you can change it to `{}` or `null` it should work the same

kit/inputs/src/UIMaterialTextView/UIMaterialTextView.tsx
```diff
@@ -62,19 +59,36 @@ function useExtendedProps(
...
+            imperativeChangeText(text, {
+                shouldSetNativeProps: false,
+            }),
+        [imperativeChangeText],
     );
 
-    const clear = useClear(resetInputHeight, onChangeText, ref);
+    React.useEffect(() => {
+        if (defaultValue) {
+            imperativeChangeText(defaultValue);
```
savelichalex: Hm, why it's done with useEffect? 🤔 isn't it better to do on the first render (with ref as a guard) and put it as a `defaultValue` to nested `TextInput`?

kit/inputs/src/UIAmountInput/hooks.ts
SAnatoliiS: In fact, this should never happen. Because UIMaterialTextView does not allow you to enter incorrect values. This condition is just insurance

savelichalex: oh really, I didn't understand that from the code :D could you please leave it as a comment in the code?

kit/inputs/src/UIMaterialTextView/UIMaterialTextView.tsx
SAnatoliiS: We need to call imperativeChangeText on first the first render to validate defaultValue with mask and to calculate other fields in this component (e.g. `checkInputHasValue` and dependent on it).
Using useEffect with empty dependency array - is the simpliest way to do this on first render.

kit/inputs/src/UIAmountInput/hooks.ts
SAnatoliiS: I reimplemented better :)

kit/inputs/src/UIMaterialTextView/UIMaterialTextView.tsx
savelichalex: But how do you guarantee a consistency between mount and `useEffect`?

savelichalex: Because between those two states input will have an incorrect value

SAnatoliiS: Why I need guarantee this consistency? Update takes tiny time. I don't think we need to make it less. But this way guarantee ref to be assigned, that needed to proper work of this component.

savelichalex: 1. useEffect doesn't guarantee that ref will be assigned.
2. on mount you have an incorrect `defaultValue`, that you try to override, it might blink to user, that leads to bad UI
3. how do you guarantee that it takes "tiny time"? I mean, that you tested it on your powerful machine doesn't make it very fast

SAnatoliiS: I can use useLayoutEffect to make this calculations before first render without blink.
Putting defaultValue to nested TextInput is wrong because first of all defaultValue should be validated by mask.
This is done by imperativeChangeText.

savelichalex: `useLayoutEffect` doesn't make calculations before a first render, in fact it actually can fire after `useEffect` :)

SAnatoliiS: Done

savelichalex (approved):

PR #1550, opened 2022-03-28
Remove details table

vetkamak:


sertony (approved):
I actually expected we'd implement these rows in Surf itself using `createUIListItemRenderer` function, but I think that it's actually OK since such table rows are used in lots of places and having them inside of casts is not a bad idea.

@savelichalex FYI

Example/src/screens/Inputs.tsx
```diff
@@ -269,18 +199,6 @@ export const Inputs = () => {
                     />
                 </View>
             </ExampleSection>
-            <ExampleSection title="UITransferInput">
```
savelichalex: Isn't it provide a conflict with PR from @SAnatoliiS ? 🤔

casts/rows/src/UITableRow/UITableRow.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { UILinkButton } from '@tonlabs/uikit.controls';
+import { UILayoutConstant, UISkeleton } from '@tonlabs/uikit.layout';
+
+import { UITableRowValueKind } from './types';
+import type { UITableRowValue, UITableRowProps } from './types';
+
+function renderTableValue(value: UITableRowValue) {
+    const { testID, kind, props } = value;
+
+    switch (kind) {
```
savelichalex: I'm not sure that I like nested `kind`s, what about make it "plain"? Also do we really need it in UIKit, maybe better to keep it in Surf for particular place only, to not have arbitrary `props`?

kit/themes/src/UILabel/index.tsx
```diff
@@ -4,3 +4,6 @@ import { ColorVariants } from '../Colors';
 export * from './UILabel';
 export const UILabelRoles = TypographyVariants;
 export const UILabelColors = ColorVariants;
+
+import type { Props } from './types';
```
savelichalex: What do you think to rather rename `Props` to `UILabelProps` in place, to have proper `Cmd+Click` behaviour?

casts/rows/src/types.ts
```diff
@@ -1,20 +1,23 @@
...
 }
 
 export type UIListRow<Kind, Props, Payload = void> = Payload extends void
     ? { kind: Kind; props: Props; key: string }
     : { kind: Kind; props: Props; key: string; payload: Payload };
 
 export type UIKitListRow<P = void> =
-    | UIListRow<UIKitListRowKind.Link, UILinkProps, P>
+    | UIListRow<UIKitListRowKind.Account, UIAccountRowProps, P>
     | UIListRow<UIKitListRowKind.Currency, UICurrencyRowProps, P>
```
savelichalex, line 17: So now we have two things to use for `currency`. That's exactly why I don't like it to be in UIKit :) How one should choose proper implementation now?

Example/src/screens/Inputs.tsx
sertony: Maybe :) Someone will have to resolve it)

casts/rows/src/UITableRow/UITableRow.tsx
sertony: AFAIK, the plain rows look very differently from the design of "Table" related rows...

casts/rows/src/types.ts
vetkamak: They have different appearance and use cases:
- Currency row has icon and caption (except for the title and amount) and should be used for a particular currency or token representation;
- Table row is simpler in appearance and can be used to display a list of fields of different nature (they are quite common is Surf);

As it is quite common and frequently using thing it seemed to me that including it into the UIKit system would be better neither adding some new component separately in Surf

kit/themes/src/UILabel/index.tsx
vetkamak: Good point, I'll change it :)

casts/rows/src/types.ts
sertony: Hm... OK, sounds logical.

@vetkamak do you think it's gonna be quite fast for you to move these table rows implementation into Surf and use `createUIListItemRenderer` method instead as it's done for `TransactionDetailsList`?

sertony (approved):

SAnatoliiS (approved):

savelichalex (approved):

PR #1554, opened 2022-03-29
Fix UISeedPhraseTextView behaviour.

SAnatoliiS:


casts/addressText/src/UIAddressTextView.tsx
```diff
@@ -31,12 +31,16 @@ type UIAddressTextViewProps = UIMaterialTextViewProps & {
     };
 };
 
-function useAddressTextView(ref: React.Ref<TextInput> | null, props: UIAddressTextViewProps) {
+function useAddressTextView(props: UIAddressTextViewProps) {
+    /**
+     * ref is passed as null because types of ref are incompatible
+     * and we don't need the clear method
```
savelichalex, line 37: Are you sure we don't need it? 🤔

kit/inputs/src/UIMaterialTextView/hooks/useImperativeChange.ts
```diff
@@ -35,16 +30,9 @@ export function useImperativeChange(
                 });
             }
 
-            if (multiline) {
-                if (Platform.OS === 'web') {
-                    const elem = ref.current as unknown as HTMLTextAreaElement;
-                    calculateWebInputHeight(elem);
```
savelichalex, line 41: So, it doesn't work anymore or why it's dropped?

casts/addressText/src/UIAddressTextView.tsx
SAnatoliiS: clear method returns from useUITextViewValue but it doesn't take here.

kit/inputs/src/UIMaterialTextView/hooks/useImperativeChange.ts
SAnatoliiS: It was duplicate logic. It works in useAutogrow.web.ts.

savelichalex (approved):

sertony (approved):

PR #1562, opened 2022-04-06
8.0 Fixes

savelichalex:
There're 2 fixes, sorry:
- one is about lagging opening of sheets
- second is about fixing auto insets calculation (again)

### DIFF
--- a/kit/scrolls/ios/UIKitScrollView.m
+++ b/kit/scrolls/ios/UIKitScrollView.m
@@ -52,6 +52,12 @@
     }
 }
 
+- (void)layoutSubviews {
+    [super layoutSubviews];
+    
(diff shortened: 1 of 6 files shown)


kit/scrolls/ios/UIKitScrollView.m
```diff
@@ -52,6 +52,12 @@ - (void)didMoveToWindow {
     }
 }
 
+- (void)layoutSubviews {
```
savelichalex, line 55: This is actual fix of auto insets calculation, as just to measure them on mount isn't enough, we have to take layout changes into account too

sertony: 👍🏻

sertony (approved):

PR #1565, opened 2022-04-07
Change keyboard type for amount input.

SAnatoliiS:


savelichalex (approved):

kit/inputs/src/UIAmountInput/UIAmountInput.tsx
```diff
@@ -37,7 +37,7 @@ export const UIAmountInput = React.forwardRef<UIAmountInputRef, UIAmountInputPro
                 success={success}
                 mask={mask}
                 onChangeText={onChangeText}
-                keyboardType="decimal-pad"
+                keyboardType="numbers-and-punctuation"
```
sertony, line 40: Note that according to the documentation this works only for iOS:
https://reactnative.dev/docs/textinput#keyboardtype

Otherwise it should fallback to `default`.

sertony: If you try to apply the fix for Android device, it helps rather because this keyboard type switches a device into the `default` keyboard...

sertony: besides iOS is getting worser with it )

PR #1568, opened 2022-04-11
Update address input QR icon & icon tap zone

vetkamak:


savelichalex (approved):

kit/layout/src/UILayoutConstant.ts
```diff
@@ -4,6 +4,7 @@ const GRID_LINE_HEIGHT = 4;
 
 export const UILayoutConstant = {
     iconSize: 24,
+    tapZoneSize: 48,
```
SAnatoliiS: How much do we need this global parameter? It seems to me that the size of the tap zone should depend on the size of the component. It seems to me it is better to define this constant locally, in the folder of this component.
What do you think @savelichalex ?

kit/inputs/src/UIMaterialTextView/hooks/useMaterialTextViewChildren.tsx
```diff
@@ -208,6 +208,12 @@ export function useMaterialTextViewChildren(
 }
 
 const styles = StyleSheet.create({
+    iconTapZone: {
+        height: UILayoutConstant.tapZoneSize,
+        width: UILayoutConstant.tapZoneSize,
```
SAnatoliiS: By setting the dimensions, you changed the layout of the components (because of this changes space between icon was increased). Tap zones should not change the layout.

sertony (commented):
I agree with @SAnatoliiS's concerns

kit/layout/src/UILayoutConstant.ts
savelichalex: yeah, actually it might depends

kit/inputs/src/UIMaterialTextView/hooks/useMaterialTextViewChildren.tsx
```diff
@@ -208,12 +216,18 @@ export function useMaterialTextViewChildren(
...
+        justifyContent: 'center',
+    },
     iconSize: {
         width: UILayoutConstant.iconSize,
         height: UILayoutConstant.iconSize,
     },
-    iconsFiller: {
-        marginLeft: UILayoutConstant.smallContentOffset,
+    iconOffset: {
+        marginRight: -UILayoutConstant.normalContentOffset,
```
sertony, line 230: I'm a bit confused with the negative offset... but will approve it)

@SAnatoliiS @savelichalex what do you think?

sertony (approved):

PR #1571, opened 2022-04-12
Do not hide keyboard on clear

SAnatoliiS:


kit/inputs/src/UIMaterialTextView/hooks/useClear.ts
```diff
@@ -11,7 +11,7 @@ export function useClear(
         function clear() {
             imperativeChangeText('');
             ref.current?.clear();
-            ref.current?.blur();
+            ref.current?.focus();
```
sertony, line 14: Btw, is it really needed?
For instance I could have a prefilled text input with a "Clear" button shown (not focused). Should I focus such input when pressing the "Clear" button? (Probably should, I don't remember the best UX practice here)

sertony (approved):

savelichalex (approved):

kit/inputs/src/UIMaterialTextView/hooks/useClear.ts
SAnatoliiS: In my opinion, yes, there should be a focus, because you pressed the button inside the input

savelichalex: Oh, while I remember, it might be a good idea to add a `focus` feature on press of the whole box around input, it's just with the new input I misclicked few times when I used Surf and was surprised that it didn't set a focus to an input, until I realized that I had to tap on the input itself :)

sertony: Discussed in PM

PR #1591, opened 2022-04-26
Do not show Expand button in UIExpandableText when text is fitted.

SAnatoliiS:


casts/texts/src/UIExpandableText/UIExpandableText.tsx
```diff
@@ -5,26 +5,29 @@ import { TouchableOpacity } from '@tonlabs/uikit.controls';
...
     const [isExpanded, setIsExpanded] = React.useState<boolean>(false);
+    const [isFit, setIsFit] = React.useState<boolean>(false);
 
     const onExpand = React.useCallback(() => {
         setIsExpanded(true);
     }, []);
 
     return (
         <View style={styles.container} testID={testID}>
+            <MeasureLabel {...props} setIsFit={setIsFit} />
```
savelichalex: I wonder why you used separate `Text` component to render it, is it because of `numberOfLines`? 🤔

SAnatoliiS: Yes, in order to render the full text, measure its dimensions and find out the number of lines. I didn't find any other way to find out the number of rows on iOS and web. If you know any - please share.

savelichalex: I might know, but not sure that it's easier :)

1. I wonder if it's possible to look at 3 dots at the end, as an indicator that text was truncated
2. There're some native APIs, it sth that I dream of for a while, but it definitely not an easy thing to do (and yep, it lacks web support)

I think your solution is ok, just please leave a comment about `numberOfLines` to explain why you decided to do it this way

SAnatoliiS: I left a comment about `MeasureLabel` component.

savelichalex (approved):

sertony (approved):

PR #1599, opened 2022-05-11
Material text view do not show clear button with children

SAnatoliiS:


savelichalex (approved):

kit/inputs/src/MaterialTextView/MaterialTextViewSimple.tsx
```diff
@@ -0,0 +1,58 @@
...
+        return (
+            <MaterialTextViewComment {...props}>
+                <View
+                    style={styles.container}
+                    // @ts-expect-error
+                    onMouseEnter={onMouseEnter}
+                    onMouseLeave={onMouseLeave}
+                    ref={borderViewRef}
+                >
+                    <Animated.View style={styles.input} /* layout={Layout} */>
```
sertony, line 27: What about `layout`?

savelichalex: Btw, I read changelogs from the latest reanimated update on Fabric, and it seems LayoutAnimations still isn't migrated, that makes them even more worse to use than before

sertony (approved):

PR #1605, opened 2022-05-16
Fix multiline text in web.

SAnatoliiS:


savelichalex (approved):

stories/chats/src/BubbleTransaction.tsx
```diff
@@ -205,7 +205,6 @@ export function BubbleTransaction(props: TransactionMessage) {
                             {...props.comment}
                             status={props.status}
                             type={props.info.type}
-                            onPress={props.onPress}
```
SAnatoliiS, line 208: `UIPressableArea` use onPress prop.

sertony (approved):

PR #1608, opened 2022-05-18
Add patch-package.

SAnatoliiS:


savelichalex (approved):
I'm not actually happy that it doesn't work with `npx`, but if we don't have another choice...

sertony (approved):

PR #1610, opened 2022-05-18
Move more input handlers to UI thread

savelichalex:
So I tried to fix a bug, when you tap a clear button and an input mess up with it's internal state.
It happened because `onBlur` actually can fire with a delay. Though an easy fix was to just put `rAF` around `clear` call in the button, but I already wanted to move more logic to UI thread, moving `onBlur` to UI thread, it fires much faster than before, basically removing the bug I mentioned.

It worth to mention that it's actually not a complete solution (for example it still updates text via a ref, though we can apply mask directly from UI thread), I would rather call it a first step, before we do sth truly cool, and we have to discuss it internally with @SAnatoliiS

### DIFF
--- a/kit/inputs/src/useTextViewHandler.ts
+++ b/kit/inputs/src/useTextViewHandler.ts
@@ -1,0 +1,63 @@
+import type {
+    TextInputChangeEventData,
+    TextInputFocusEventData,
+    TextInputSelectionChangeEventData,
+} from 'react-native';
+import { useEvent, useHandler } from 'react-native-reanimated';
+
+export function useTextViewHandler<Context extends Record<string, unknown>>(
+    handlers:
+        | {
+              onFocus?: (evt: TextInputFocusEventData, ctx: Context) => void;
+              onBlur?: (evt: TextInputFocusEventData, ctx: Context) => void;
+              onChange?: (evt: TextInputChangeEventData, ctx: Context) => void;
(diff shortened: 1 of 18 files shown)


kit/inputs/src/useTextViewHandler.ts
```diff
@@ -0,0 +1,63 @@
...
+    const subscribeForEvent = ['topChange'];
+    if (textViewHandlers.onFocus != null) {
+        subscribeForEvent.push('topFocus');
+    }
+    if (textViewHandlers.onBlur != null) {
+        subscribeForEvent.push('topBlur');
+    }
+    if (textViewHandlers.onSelectionChange != null) {
+        subscribeForEvent.push('topSelectionChange');
+    }
```
sertony, line 34: This is not documented. Have you found it at RN repo code?
https://github.com/facebook/react-native/blob/d0871d0a9a373e1d3ac35da46c85c0d0e793116d/React/Base/RCTEventDispatcher.m#L24

savelichalex: Not exactly :D But anyway you right, it is what you found. I actually just found a place where `reanimated` captures them and put a log there :D

sertony (approved):
This is a rocket!

SAnatoliiS (approved):

PR #1613, opened 2022-05-20
fix folded accordion section double-clicking to open

vetkamak:


kit/scrolls/src/Lists/UIAccordionSectionList/UIAccordionSectionList.tsx
```diff
@@ -222,9 +222,21 @@ function UIAccordionSectionListInner<ItemT>({
...
+        sections.forEach(section => {
+            const { key, isFolded } = section;
+            const sectionKey = `${key}:header`;
+            if (isFolded && !(sectionKey in foldedSections)) {
+                setFoldedSections({
+                    ...foldedSections,
+                    [sectionKey]: true,
+                });
+            }
+        });
```
savelichalex: Can you please explain how exactly does it fix the problem? 🤔 
I see that there is already a logic for that https://github.com/tonlabs/UIKit/pull/1613/files#diff-9603277eb9363ebb5ad83739f59f6641f63233d3311873559ee81b916ac2a902R244-R252
how does it differ from that?

vetkamak: It fixes the initial state of foldedSections - we start mentioning isFolded prop only after first click on initially folded section.

savelichalex: Ok, I see few problems here:
1. It triggers n re-renders (initial render, and then n re-renders based on amount of `setFoldedSections` calls)
2. It's going to run on every sections change (not only once after mount)
3. It's going to run the second time, after you change `foldedSections` (second argument to useEffect will be changed) (even though it won't call `setFoldedSection`, so it's just a waste of CPU resources)
4. The logic I [mentioned](https://github.com/tonlabs/UIKit/pull/1613/files#diff-9603277eb9363ebb5ad83739f59f6641f63233d3311873559ee81b916ac2a902R244-R252), that is actually about the same, either one of them must be deleted

savelichalex (approved):

kit/scrolls/src/Lists/UIAccordionSectionList/UIAccordionSectionList.tsx
```diff
@@ -219,33 +253,26 @@ function UIAccordionSectionListInner<ItemT>({
...
+                isFolded: foldedSections[getVirtualizedHeaderKey(key)],
             };
         });
     }, [sections, foldedSections]);
 
+    // This is used to reduce re-creation of a press callback
+    const processedSectionsHolderRef = React.useRef<typeof processedSections>(processedSections);
+    React.useEffect(() => {
+        processedSectionsHolderRef.current = processedSections;
+    }, [processedSections]);
```
sertony: Don't see why we need it...

sertony (approved):

kit/scrolls/src/Lists/UIAccordionSectionList/UIAccordionSectionList.tsx
```diff
@@ -219,33 +253,26 @@ function UIAccordionSectionListInner<ItemT>({
...
+
```
savelichalex:
```suggestion
```

PR #1614, opened 2022-05-20
Transfer UIPillButton to Pressable component.

SAnatoliiS:


sertony (approved):

kit/controls/src/UIPillButton/PillButtonContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    const animatedBackgroundOverlayStyle = useAnimatedStyle(() => {
+        return {
+            backgroundColor: backgroundOverlayColor.value,
+        };
+    });
+
+    React.useLayoutEffect(() => {
+        LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut);
+    }, [loading]);
```
savelichalex, line 61: hm, why do we need it? 🤔

SAnatoliiS: To animate button shape transition between loading and regular state on mobile platform.

savelichalex: Could you please add it as a comment in the code 🙏 

PS. I'm a bit suspicious to that as it might not work well (mixing reanimated and LayoutAnimations from RN), but AFAIK  it worked fine in this case

savelichalex (approved):

PR #1618, opened 2022-05-25
Add representation of disabled state of MaterialTextView.

SAnatoliiS:


kit/inputs/src/MaterialTextView/MaterialTextView.tsx
```diff
@@ -97,6 +99,10 @@ function useExtendedProps(
         onFocus: evt => {
             'worklet';
 
+            if (editable.value === false) {
+                return;
+            }
+
```
savelichalex, line 108: wait, even though input isn't editable it still fire focus/blur events??? This is weird

savelichalex (approved):

kit/inputs/src/MaterialTextView/MaterialTextView.tsx
SAnatoliiS: This behavior is only on web

savelichalex: Please add a comment, need to think how to avoid it for mobiles 🤔

SAnatoliiS: Or maybe fix this behavior in `useTextViewHandler` for web.

savelichalex: yep, that might work, like for example doesn't return `onFocus` and `onBlur`

sertony (approved):

PR #1621, opened 2022-05-26
Add icon to browser menu.

SAnatoliiS:


stories/chats/src/BubbleActionButton.tsx
```diff
@@ -41,13 +51,26 @@ export function BubbleActionButton(message: ActionButtonMessage) {
...
     }, [variant]);
 
+    const iconPosition = React.useMemo<UIMsgButtonIconPosition | undefined>(() => {
+        switch (iconPositionProp) {
+            case ActionButtonMessageIconPosition.Left:
+                return UIMsgButtonIconPosition.Left;
+            case ActionButtonMessageIconPosition.Right:
+                return UIMsgButtonIconPosition.Middle;
+            default:
+                return undefined;
```
savelichalex: Maybe provide a default position?

stories/chats/src/BubbleActionButton.tsx
```diff
@@ -41,13 +51,26 @@ export function BubbleActionButton(message: ActionButtonMessage) {
...
 
+    const iconPosition = React.useMemo<UIMsgButtonIconPosition | undefined>(() => {
+        switch (iconPositionProp) {
+            case ActionButtonMessageIconPosition.Left:
+                return UIMsgButtonIconPosition.Left;
+            case ActionButtonMessageIconPosition.Right:
+                return UIMsgButtonIconPosition.Middle;
+            default:
+                return UIMsgButtonIconPosition.Left;
+        }
```
sertony, line 62: Smth is wrong here, no?
If it's right, why should it be middle?

SAnatoliiS: Middle - it is right position from title and up from caption. Because there is no caption so it's ok.

savelichalex (approved):

sertony (approved):

PR #1636, opened 2022-06-10
add color and export Pressable

vetkamak:


savelichalex:
@SAnatoliiS what do you think?

kit/controls/src/index.ts
```diff
@@ -26,6 +26,7 @@ export * from './UIShowMoreButton';
 export * from './UISwitcher';
 export * from './UIWideBoxButton';
 export * from './UIPressableArea';
+export * from './Pressable';
```
SAnatoliiS, line 29: `Pressable` is an internal module. It is not worth exporting, because it has its own peculiarities of use. In order to make the area tappable, you can use a ready-made `UIPressableArea` component.

sertony: @SAnatoliiS it's just we need to reimplement the button completely in Surf since @eteslov provided a new design)) That's why @vetkamak had to use "private" methods ;)

kit/themes/src/Colors.ts
```diff
@@ -92,6 +93,7 @@ const StaticTheme = {
     [ColorVariants.StaticBackgroundNegative]: 'rgba(231, 23, 23, 0.12)',
     [ColorVariants.StaticBackgroundPositive]: 'rgba(54, 192, 92, 0.12)',
     [ColorVariants.StaticBackgroundOverlay]: 'rgba(4, 4, 4, 0.4)',
+    [ColorVariants.StaticBackgroundSurfPrimary]: '#4963E6',
```
sertony: @savelichalex @vetkamak what do you think to keep this color only in Surf for now? (i.e. just export Pressable)

kit/controls/src/index.ts
SAnatoliiS: @sertony  Are you really sure that you need to create buttons in `Surf` and not in `UIKit`?

sertony: @SAnatoliiS , this is @eteslov who wants it there ASAP =) We'll try to move them (or reimplement) in UIKit as well soon

savelichalex: What do you guys think about to move it to UIKit? Shoudn't be hard to do. What cons do you see?

SAnatoliiS (approved):

kit/controls/src/index.ts
vetkamak: I suppose there is no need in moving it to the UIKit - if I correctly understand we will update colors and current buttons soon and added Button will be removed after that. So it is a kind of temporary solution in order to fit the design, after updates we can turn Pressable back to an internal one

sertony (approved):

PR #1650, opened 2022-07-07
Highlights

savelichalex:


casts/highlights/src/UIHighlights.tsx
```diff
@@ -0,0 +1,301 @@
...
+                        onMomentumScrollBegin: () => undefined,
+                        flingEnabled: !pagingEnabled,
+                    },
+                })}
+                {...onWheelProps}
+            >
+                {React.Children.map(children, (child, itemIndex) => {
+                    return (
+                        <View
+                            style={
```
SAnatoliiS: Should we add the 'key' prop here?

SAnatoliiS (approved):

casts/highlights/src/UIHighlights.tsx
savelichalex: Good question, let me see if it's possible to gather it from children

PR #1652, opened 2022-07-15
Pass onClose if it's needed for a modal route

savelichalex:


SAnatoliiS (approved):

casts/modalNavigator/src/ModalNavigator/createModalNavigator.tsx
```diff
@@ -34,7 +34,7 @@ function ModalScreen<ParamList extends ParamListBase = ParamListBase>({
         <UIModalSheet
             forId="modal"
             visible={route.visible}
-            onClose={hide}
+            onClose={'onClose' in route ? route.onClose : hide}
```
BigRoma: It's too much. What I must do if I need only handle event but need close modal anyway?
Suggest transfer onClose event to hide function.

BigRoma (changes requested):

sertony (approved):

PR #1658, opened 2022-07-29
Navigation perf profiling

savelichalex:


Example/android/app/src/main/java/tonlabs/uikit/MainApplication.java
```diff
@@ -35,7 +37,11 @@ protected List<ReactPackage> getPackages() {
           // Packages that cannot be autolinked yet can be added manually here, for example:
           // packages.add(new MyReactNativePackage());
 
-          return packages;
+            packages = PatchedReanimatedPackage.patchPackages(packages);
```
savelichalex, line 40: It might be a good idea to apply it in Surf anyway as it completely remove patched UIManager from reanimated that they use for layout animations and it has useless overhead

PR #1669, opened 2022-08-26
Small layout updates

vetkamak:


sertony (approved):

casts/highlights/src/UIHighlights.tsx
```diff
@@ -213,98 +213,101 @@ export function UIHighlights({
...
-                            return (
-                                <View
-                                    key={child.key}
-                                    style={
-                                        itemIndex !== 0
-                                            ? {
-                                                  paddingLeft: spaceBetween,
-                                              }
-                                            : null
+        React.Children.count(children)
```
sertony: Can you comment why this was needed?

casts/highlights/src/UIHighlights.tsx
```diff
@@ -213,98 +213,101 @@ export function UIHighlights({
...
+            ? (
```
SAnatoliiS: I suggest using a separate `if` operator instead of this huge ternary operator

sertony (approved):

PR #1673, opened 2022-09-09
Add new notice

SAnatoliiS:
image
image


kit/popups/src/UIInteractiveNotice/Actions.tsx
```diff
@@ -0,0 +1,49 @@
...
+    return <View style={styles.container}>{actionList.map(Action)}</View>;
+}
+
+const styles = StyleSheet.create({
+    container: {
+        flexDirection: 'row',
+        flexWrap: 'wrap',
+        marginTop: UILayoutConstant.contentInsetVerticalX1,
+        marginHorizontal: -UILayoutConstant.contentOffset / 2,
+        marginBottom: -UILayoutConstant.contentInsetVerticalX2,
```
sertony, line 43: Why do we have to use negative margins here?

kit/popups/src/UIInteractiveNotice/CloseButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+    if (!showCloseButton) {
+        return null;
+    }
+    return (
+        <View style={style}>
+            <UIPressableArea
+                onPress={onClose}
+                style={{
+                    marginVertical: -UILayoutConstant.contentInsetVerticalX4,
+                    marginHorizontal: -UILayoutConstant.contentOffset,
```
sertony, line 24: Same question. Besides, does it work well with Android where all the Views have overflow: 'hidden'?

sertony (approved):

kit/popups/src/UIInteractiveNotice/Actions.tsx
SAnatoliiS: To make a large tap zone without remake all padding layout principles.

kit/popups/src/UIInteractiveNotice/CloseButton.tsx
SAnatoliiS: It works well on all platforms, I checked.

vetkamak (approved):

PR #1682, opened 2022-10-05
Add new seed phrase input

SAnatoliiS:


sertony (approved):
Good. Thanks. @vetkamak should also check it.

I also believe that @eteslov should have add some kind of hints in the design, otherwise it's getting difficult to work it when typing the phrase ;)

vetkamak (approved):

PR #1689, opened 2022-10-28
Fix RTL tab bar dots

SAnatoliiS:


Example/babel.config.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
     presets: ['module:metro-react-native-babel-preset'],
     plugins: [
+        '@babel/plugin-proposal-export-namespace-from',
```
sertony, line 4: Oh, true. We did it in Surf as well

sertony (approved):

PR #1695, opened 2022-11-23
Fix rtl highlights

SAnatoliiS:


casts/highlights/src/RtlConverter.ts
```diff
@@ -0,0 +1,23 @@
+import { Platform } from 'react-native';
+
+const { OS } = Platform;
+
+export function normalizeScrollX(
```
sertony, line 10: Can you describe why we need this function and why it's so platform dependent?

casts/highlights/src/useOnWheelHandler.ts
```diff
@@ -1,13 +1,33 @@
 /* eslint-disable no-param-reassign */
 import * as React from 'react';
+import { I18nManager, NativeScrollEvent, NativeSyntheticEvent } from 'react-native';
 import { nanoid } from 'nanoid/non-secure';
 
 import { getWorkletFromParentHandler } from '@tonlabs/uikit.scrolls';
-import type { NativeScrollEvent, NativeSyntheticEvent } from 'react-native';
+import { clamp } from 'lodash';
```
sertony:
```suggestion
import clamp from 'lodash/clamp';
```

in order to make the final bundle smaller

casts/highlights/src/UIHighlights.tsx
```diff
@@ -109,53 +74,69 @@ export function UIHighlights({
...
                 return;
             }
 
-            // Check if we've reached the end
-            if (x >= scrollViewContentWidthRef.current - scrollViewWidthRef.current) {
-                // Set the current gravity position as for the last item
-                currentGravityPosition.value = sharedContext.value.positions.length - 1;
-                currentProgress.value = 0;
-                return;
-            }
```
sertony, line 158: AFAIR this was very important fix to show the dot when the last card could not be scrolled to the left border of the visible view.

Don't we need it anymore?

casts/highlights/src/useOnWheelHandler.ts
```diff
@@ -39,18 +64,28 @@ function wrapVerticalToHorizontalWheelMovementHandler(
         // don't propagate to an outter scroll view if any
         event.stopPropagation();
 
-        const deltaCoord = event.deltaY || event.deltaX;
-        // Note: e.deltaY is not present for `DOMMouseScroll` event (used by Firefox)
```
sertony, line 43: I wonder if this is still valid for Firefox :) Can you check?

casts/highlights/src/UIHighlights.tsx
SAnatoliiS: You are right. I missed this. I added this logic to [usePosition.tsx](https://github.com/tonlabs/UIKit/pull/1695/files#diff-9f4515115482259d6c48fe751437cd954d109a84db48dd2b576519e889abbdbaR113-R123)

casts/highlights/src/useOnWheelHandler.ts
SAnatoliiS: I checked on my Firefox browser. There is a regular [wheel](https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event) event.

sertony (approved):

casts/highlights/src/useOnWheelHandler.ts
sertony: Is the speed of it actually OK? As I see the former hack was applied by multiplying the delta to some factor…

SAnatoliiS: I checked on Chrome, Firefox and Safari. The speed is ok.

PR #1698, opened 2022-11-24
Fix tooltip position

SAnatoliiS:


kit/popups/src/useWindowDimensions.web.ts
```diff
@@ -0,0 +1,11 @@
+import { useDimensions } from '@tonlabs/uicast.keyboard';
```
sertony: What do you think of moving this hook into another package? It doesn't look like it should belong to the .keyboard

SAnatoliiS: I agree. Done.

sertony (approved):

PR #1702, opened 2022-12-07
Fix RTL input text position.

SAnatoliiS:


kit/inputs/src/MaterialTextView/hooks/useClear.ts
```diff
@@ -9,7 +9,7 @@ export function useClear(
     return React.useCallback(
         function clear() {
             imperativeChangeText('');
```
sertony, line 11: Does this call ensure that the same effect as `ref.current?.clear();` is applied? I just want to say if we need to remove the `ref.current?.clear();` line bellow?

kit/inputs/src/UIAmountInput/hooks.tsx
```diff
@@ -148,9 +148,13 @@ export function useUIAmountInputChildren(
...
+    if (hideClearButton) {
+        return undefined;
+    }
+
+    if (inputHasValue && (isFocused || isHovered)) {
         return <MaterialTextViewClearButton clear={clear} />;
     }
 
-    return undefined;
+    return <MaterialTextViewClearButton hiddenButton />;
```
sertony: I suggest to add comments for each conditions in order to recall once we need it, why the logic for this component is this one. Since it differs to the logic of UIMaterialTextView component

kit/inputs/src/MaterialTextView/types.ts
```diff
@@ -120,7 +120,8 @@ export type MaterialTextViewTextProps = {
 };
 
 export type MaterialTextViewClearButtonProps = {
-    clear: (() => void) | undefined;
+    clear?: (() => void) | undefined;
+    hiddenButton?: boolean;
```
sertony, line 131: Let's also add a comment why do we need it (I'd also add a TSDoc comment for each prop here)

kit/inputs/src/MaterialTextView/hooks/useClear.ts
SAnatoliiS: I didn't notice any difference in behavior of `ref.current?.clear()` and `imperativeChangeText('')`. But for inner logic it better to use imperativeChangeText on every text change (e.g. for autogrow of the input on web).

sertony (approved):

PR #1713, opened 2022-12-28
Flip images in RTL mode.

SAnatoliiS:


kit/media/src/UIImage/hooks/useFlippedImageIfRtl.ts
```diff
@@ -0,0 +1,31 @@
...
+export function useFlippedImageIfRtl(
+    style: StyleProp<ImageStyle> | undefined,
+): StyleProp<ImageStyle> | undefined {
+    const isRTL = React.useMemo(() => I18nManager.getConstants().isRTL, []);
+
+    const flipImageIfRtlStyle: StyleProp<ImageStyle> = isRTL
+        ? {
+              transform: [
+                  {
+                      rotateY: '180deg',
```
sertony: If the purpose is to flip the image horizontally than it should be:
```suggestion
                      scaleX: -1,
```

PR #1716, opened 2023-01-09
Get rid of core package

SAnatoliiS:


casts/pinCode/src/UIPinCode/BlockingView.tsx
```diff
@@ -17,7 +17,7 @@ function BlockingViewInner({ onBlockingViewReady, onBlockingViewGone }: Blocking
         onBlockingViewReady();
 
         return onBlockingViewGone;
-    }, []);
+    }, [onBlockingViewGone, onBlockingViewReady]);
```
sertony, line 20: 👍

sertony (approved):
Great PR! 🔥 

Did you run `eslint:fix` and `prettier:fix` before committing?
If they run successfully I strongly suggest to add "eslint:check" & "prettier:check" in the "pre-commit" of the package.json! (We do it in another project)

SAnatoliiS:
> Great PR! 🔥
> 
> Did you run `eslint:fix` and `prettier:fix` before committing? If they run successfully I strongly suggest to add "eslint:check" & "prettier:check" in the "pre-commit" of the package.json! (We do it in another project)

Yes I did and they run successfully.
I will do what you suggested.

sertony (approved):

PR #1718, opened 2023-01-13
Fix debot flipped text.

SAnatoliiS:


stories/chats/src/BubblePlainText.tsx
```diff
@@ -1,18 +1,10 @@
...
-import {
-    TouchableWithoutFeedback,
-    StyleSheet,
-    Platform,
-    View,
-    Animated,
-    TextStyle,
-    I18nManager,
-} from 'react-native';
+import { StyleSheet, Platform, View, Animated, TextStyle, I18nManager } from 'react-native';
```
sertony: Animated from react-native is still used here (and as far as I see this component is transforming which is more likely to lead to the bug)

SAnatoliiS: Completely removed

sertony (approved):

PR #1761, opened 2023-04-27
add uicurrency signBeforeNumberNeedsNBSP prop

vetkamak:


casts/numbers/src/Numbers/UIAnimatedNumber.tsx
```diff
@@ -36,10 +36,11 @@ export function UIAnimatedNumber({
     // sign customization
     sign,
     signBeforeNumber = false,
+    signBeforeNumberNeedsNBSP = true,
```
sertony, line 39: The first prop was scaring me...
No I'm scared a lot about this prop...

sertony (changes requested):
Rejecting due to the scary API

